PR #5, opened 2021-09-29
Reintroduce Elasticsearch compatibility from OpenDistro datasource

Elfo404:


connorlindsey (dismissed):

pkg/opensearch/client/client.go
```diff
@@ -126,9 +127,11 @@ var NewClient = func(ctx context.Context, ds *backend.DataSourceInstanceSettings
...
 	if err != nil {
-		return nil, fmt.Errorf("opensearch version is required, err=%v", err)
+		return nil, fmt.Errorf("version is required, err=%v", err)
 	}
 
+	flavor := jsonData.Get("flavor").MustString(string(OpenSearch))
+
 	timeField, err := jsonData.Get("timeField").String()
 	if err != nil {
 		return nil, fmt.Errorf("opensearch time field name is required, err=%v", err)
```
ifrost, line 137: Should this also be:

```suggestion
		return nil, fmt.Errorf("time field name is required, err=%v", err)
```

Also at line 114: 

```diff
- return nil, fmt.Errorf("error reading opensearch version")
+ return nil, fmt.Errorf("error reading version")
```

pkg/opensearch/time_series_query_test.go
```diff
@@ -14,10 +15,48 @@ import (
 func TestExecuteTimeSeriesQuery(t *testing.T) {
 	from := time.Date(2018, 5, 15, 17, 50, 0, 0, time.UTC)
 	to := time.Date(2018, 5, 15, 17, 55, 0, 0, time.UTC)
+	fromStr := fmt.Sprintf("%d", from.UnixNano()/int64(time.Millisecond))
+	toStr := fmt.Sprintf("%d", to.UnixNano()/int64(time.Millisecond))
 
 	Convey("Test execute time series query", t, func() {
+		Convey("With defaults on Elasticsearc 2.0.0", func() {
```
ifrost, line 22:
```suggestion
		Convey("With defaults on Elasticsearch 2.0.0", func() {
```

pkg/opensearch/client/client.go
```diff
@@ -126,9 +127,11 @@ var NewClient = func(ctx context.Context, ds *backend.DataSourceInstanceSettings
 
 	version, err := extractVersion(jsonData.Get("version"))
 	if err != nil {
-		return nil, fmt.Errorf("opensearch version is required, err=%v", err)
+		return nil, fmt.Errorf("version is required, err=%v", err)
```
ifrost, line 130: When request fails with error it read "OpenSearch error..." which could be confusing is someone has configured Elastic? For example when testing datasource it shows below even when Elastic is selected:

Screen Shot 2021-10-05 at 08 53 40

ifrost (approved):

pkg/opensearch/client/client.go
Elfo404: That's true, also `undefined. undefined` is not really the best error message ever.
I'll create an issue to improve error handling in general.

ifrost (approved):

PR #9, opened 2021-11-11
Improve error handling

### LINKED ISSUE #6: Improve error handling
When request fails with error it read "OpenSearch error..." which could be confusing is someone has configured Elastic? For example when testing datasource it shows below even when Elastic is selected:

Screen Shot 2021-10-05 at 08 53 40

_Originally posted by @ifrost in https://github.com/grafana/opensearch-datasource/pull/5#discussion_r721940972_

In general we should improve error handling to have better messages.

Elfo404:
Slightly improves error handling. the logic to extract the error message is mirroring what is currently being done in the upstream Elasticsearch datasource.

Screenshot 2021-11-11 at 11 41 22

Fixes #6 

cc. @ifrost I also wanted to prefix the error based on the configured "flavor" (OpenSearch / Elasticsearch) but it felt even more confusing in the following situation:
- I have an OpenSearch cluster running
- I set Elasticsearch 7.0.0 in settings
- the error would be "Elasticsearch error: ..." while in fact I'm getting the error from an OpenSearch cluster.

also, I might even not connect at all to an OpenSearch/ES cluster (wrong URL or any network misconfiguration for instance), therefore i'd leave it like it is now, assuming "OpenSearch" in the error message refers to "OpenSearch plugin"

### DIFF
--- a/src/datasource.ts
+++ b/src/datasource.ts
@@ -106,9 +106,11 @@ export class OpenSearchDatasource extends DataSourceApi<OpenSearchQuery, OpenSea
     return getBackendSrv()
       .datasourceRequest(options)
       .catch((err: any) => {
-        if (err.data && err.data.error) {
+        if (err.data) {
+          const message = err.data.error?.reason ?? err.data.message ?? 'Unknown error';
+
           throw {
-            message: `OpenSearch error: ${err.data.error.reason}. ${err.data.error.details}`,
+            message: 'OpenSearch error: ' + message,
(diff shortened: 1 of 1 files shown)


ifrost (approved):
LGTM! 🚀 

> I also wanted to prefix the error based on the configured "flavor" (OpenSearch / Elasticsearch) but it felt even more confusing (...)

I think it makes sense and makes it consistent with Elasticsearch core plugin 👍

PR #16, opened 2022-01-28
Fix alias pattern not being correctly handled by Query Editor

### LINKED ISSUE #15: Alias not visible in panel with datasource Opensearch
**What happened:**
After switch of datasource to Opensearch, when setting up alias in the query panel after save they are not visible on the panel but only in the panel JSON. 

Example of the panel query before to save it:
Screenshot 2022-01-28 at 13 57 15

How it looks after saved it and refreshed the page:
Screenshot 2022-01-28 at 13 58 11

Example of the same in JSON:
Screenshot 2022-01-28 at 13 52 07

**What you expected to happen:** 
The alias to be visible after save and refresh of the page.

**How to reproduce it (as minimally and precisely as possible):**

Here is an example of how the query looks after saved with datasource Opensearch:
Screenshot 2022-01-28 at 13 36 10

And an example of how the query looks after saved with datasource Elasticsearch:
Screenshot 2022-01-28 at 13 34 43

**Anything else we need to know?:**

**Environment:**
Grafana version: grafana/grafana:7.5.10
Data source type & version: OpenSearch 1.0
User OS & Browser:
Grafana plugins: OpenSearch 1.1.1

Elfo404:
Fixes #15 

Alias pattern was not correctly being handled by the query editor after the introduction of PPL.

This PR changes it so it is handled correctly.

Note: most of the changes are in order to remove the duplication in PPLEditor and LuceneEditor, unifying them in a single component. this was needed to also fix a display issue with the Quey Field not being wrapped properly in a flex container.

to test this is working as expected, simply create any time series query (count will work) and try to edit the alias field. Before this change, after reloading the value would not be reflected in the query editor although its value was being properly used.


ivanahuckova (approved):
Great work! Works nicely! Also the refactor is :chefs-kiss:. Also it improves how the UI looks. LGMT.

ifrost (approved):

PR #30, opened 2022-05-25
Use latest version of grafana/ui and grafana/aws-sdk

sunker:
In a recent change in Grafana, the dependency to grafana/aws-sdk was moved from grafana/ui. This means that any plugin that support SIGV4 auth need to depend on the grafana/aws-sdk and pass a SIGV4 editor as prop to the DataSourceHttpSettings. To get details on why we're doing this change, please refer to [this](https://github.com/grafana/grafana/pull/43559) PR in grafana.  

This means we'll have to upgrade this plugin to use latest version of grafana/ui and grafana/aws-sdk and release a new major version of this plugin. Any existing version of this plugin will be incompatible with grafana 9.0.0 and later. 

My goal was to put up a PR with a working example, but since this plugin runs 7.5 of grafana packages there's simply too many necessary changes that are unrelated to the breaking change in the DataSourceHttpSettings so I'll need help. I also lack context in this domain as well as ways of testing this data source. 

This draft PR includes changes necessary for rendering SIGV4 correctly.


package.json
```diff
@@ -14,16 +14,17 @@
...
-    "@grafana/data": "7.5.7",
-    "@grafana/runtime": "7.5.7",
-    "@grafana/ui": "7.5.7",
     "react": "17.0.1",
     "react-dom": "17.0.1",
     "react-use": "13.27.0",
     "semver": "^7.3.5"
   },
   "devDependencies": {
+    "@grafana/aws-sdk": "0.0.36",
```
sunker, line 23: I believe these should be devDependencies.

src/plugin.json
```diff
@@ -27,8 +27,7 @@
     "updated": "%TODAY%"
   },
   "dependencies": {
-    "grafanaVersion": "7.5.7",
```
sunker, line 30: grafanaVersion is deprecated. Should use grafanaDependency instead

svennergr:
Hi @sunker,

if okay for you, I close this PR in favour of https://github.com/grafana/opensearch-datasource/pull/48.

Best
@svennergr

PR #43, opened 2022-06-27
Chore: update backend dependencies

svennergr:
**What this PR does / why we need it:**

Upgrades the defined go dependencies.

**Special notes for reviewers:**
I created those dependencies the following:
1. Issue a `go get -u` to update the dependencies. That command also added a few indirect dependencies, which were missing before.
2. `go mod tidy` to ensure everything is fine.
3. I then did some manual work by ordering the `indirect` and `direct` dependencies, like it is done here as well: https://github.com/grafana/grafana/blob/main/go.mod


gabor (approved):
LGTM

PR #51, opened 2022-07-08
Tech: Add workflow to automatically move issues to L&T project

svennergr:
Added a workflow to move issues to the L&T project for better visibility.

This can not really be tested unless it is on `main`.

Workflow is also in use e.g. here: https://github.com/grafana/grafana-plugin-sdk-python/blob/master/.github/workflows/add-issues-to-project.yml

### DIFF
--- a/.github/workflows/add-issues-to-project.yml
+++ b/.github/workflows/add-issues-to-project.yml
@@ -3,6 +3,8 @@ name: Auto Assign to Logs and Traces Project
 on:
   issues:
     types: [opened]
+  pull_request:
+    types: [opened]
 
 env:
   MY_GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/add-issues-to-project.yml
```diff
@@ -0,0 +1,20 @@
+name: Auto Assign to Logs and Traces Project
+
+on:
+  issues:
+    types: [opened]
```
ivanahuckova, line 5: Should this be just issues or also?
```
pull_request:
    types: [opened]
```

ivanahuckova: And if just issues, then probably this shouldn't have pull requests `Assign NEW issues and NEW pull requests to Logs&Traces project`

svennergr: Good idea! I added it to the file!

ivanahuckova (approved):
LGTM

PR #62, opened 2022-09-02
PPL: Fixed queries being made and interpreted in local timezones

svennergr:
The datasource was using timestamps in local format when requesting a query. That was changed to UTC.

Furthermore the datasource interpreted the receiving timestamps via a PPL query as local timestamps, whereas it should have interpreted them as UTC timestamps.


gabor (approved):
LGTM!

PR #67, opened 2022-09-09
[1.2.x] PPL: Fix timestamps in `table` visualisation

svennergr:
Backport for https://github.com/grafana/opensearch-datasource/pull/66.

Timestamps in table visualisation were shown as UTC. This PR will convert every property in a dataframe, that is type timestamp or datetime to the local time format.


gabor (approved):
LGTM!

Issue #73, opened 2022-10-14
Backend client does not honor custom headers


yuri-tceretian:
The backend implementation of the client parses the data source configuration on its own. It does not read the custom headers from the settings and does not configure the client to include them.
In the Elasticsearch client that is embedded to Grafana 
https://github.com/grafana/grafana/blob/9c5e459f0bcb83132f9de4d1ce38df9e7d5c77e7/pkg/tsdb/elasticsearch/elasticsearch.go#L64-L67
we use the SDK function to get the client configuration. That function supports custom headers.

svennergr:
Hi @yuri-tceretian,

a fix has been merged and released in v1.2.3 / v.2.0.3.

PR #74, opened 2022-10-17
Backend: Fix missing custom headers

svennergr:
As described in https://github.com/grafana/opensearch-datasource/issues/73 custom headers are currently not send when the datasource is in backend mode.

This PR adds a hotfix to add custom headers. Since this will be backported to OpenSearch datasource v1.2.x, I decided to not implement the "full" Grafana HTTP client but rather go with this trivial approach of setting the custom headers.

### DIFF
--- a/pkg/opensearch/client/client.go
+++ b/pkg/opensearch/client/client.go
@@ -30,6 +30,8 @@ var (
 
 // TODO: use real settings for HTTP client
 var newDatasourceHttpClient = func(ds *backend.DataSourceInstanceSettings) (*http.Client, error) {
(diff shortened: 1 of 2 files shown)


pkg/opensearch/client/client.go
```diff
@@ -30,6 +30,8 @@ var (
 
 // TODO: use real settings for HTTP client
 var newDatasourceHttpClient = func(ds *backend.DataSourceInstanceSettings) (*http.Client, error) {
+	//dsHttpOpts, err := ds.HTTPClientOptions()
```
gabor, line 33: do we need this line?

gabor (approved):
LGTM!

PR #80, opened 2022-11-04
Backend: Fix index being read from the wrong place

### LINKED ISSUE #78: Using alerts on panel not working
Hello!

Using alerts on existing panels are not working.

Version(s):
- Grafana `v8.5.13 (38d274060d)`
- OS DS Plugin `1.2.2`
- OS Version `1.3.2`

Example panel:
image
Example alert:
image
When clicking "Test rule" in "alerts tab" getting:
```
{
  ...
  "logs": [
    ...
    {
      "message": "Condition[0]: Query Result",
      "data": {
        "fromDataframe": true,
        "series": []                                          
      }
    },
    {
      "message": "Condition: Eval: false, Query Returned No Series (reduced to null/no value)",
      "data": null
    }
  ]
}
```

svennergr:
As reported in #78 alert conditions are failing for some users. Reason for this is that the index pattern was read from `DataSourceSettings.Database` but is actually stored at `jsonData.database`.

This PR fixes #78 by also reading the database from `jsonData`.

### DIFF
--- a/pkg/opensearch/client/client.go
+++ b/pkg/opensearch/client/client.go
@@ -140,8 +140,14 @@ var NewClient = func(ctx context.Context, ds *backend.DataSourceInstanceSettings
 		return nil, fmt.Errorf("time field name is required, err=%v", err)
 	}
 
+	db, err := jsonData.Get("database").String()
+	if err != nil {
+		// `jsonData.database` is optional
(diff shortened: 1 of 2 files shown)


gabor (approved):
LGTM!

PR #88, opened 2022-11-19
Ability to select order (Desc/Asc) for "raw data" metrics agreggations

lvta0909:
What?
Add ability to select the query order for "raw data" metric agreggation.

Why?
Sometimes it is needed to query the firsts X documents starting from a defined time. With the default query order as Descending it was not directly possible as it would give the lasts X documents from the ending time.
The ability to choose the order solves this limitation.
This functionality is available only to "raw data" and "raw document" metric queries.

order_example


sarahzinger:
@lvta0909 sorry for the delay in reviewing, this looks really good to me! I merged in master and made some minor changes if you'd like to take a look!

sarahzinger:
@iwysiu mind reviewing this since I added a few more commits? I can merge/release once you think it looks good.

sarahzinger:
Also I believe the build is failing because the levity check fails for all external contributors right now.

src/components/QueryEditor/MetricAggregationsEditor/SettingsEditor/index.test.tsx
```diff
@@ -68,10 +71,36 @@ describe('Settings Editor', () => {
...
+      let select = (await screen.findByText('Order')).nextSibling!.firstChild!;
+      await fireEvent.keyDown(select, { keyCode: 40 });
+      const scs = screen.getAllByLabelText('Select option');
+      expect(scs).toHaveLength(2);
+
+      // Define new value and trigger a click to update metric
+      const newOrderValue = 'asc';
+      await userEvent.click(screen.getByText('Ascending'));
+      expect(onChange).toHaveBeenCalledTimes(2);
+      //      expect(onChange).toHaveBeenCalledWith(expect.objectContaining({'order' : 'asc'}));
```
iwysiu, line 88: should this be removed/uncommented?

src/components/QueryEditor/MetricAggregationsEditor/SettingsEditor/index.tsx
```diff
@@ -28,6 +28,8 @@ interface Props {
   previousMetrics: MetricAggregation[];
 }
 
+const orderLabel = 'Test Order select Container';
```
iwysiu, line 31: I don't think this needs to be a separate variable

iwysiu (commented):
Beyond a couple style comments, nice! Though as a note, do we want `desc` to be in the label when the order isn't settable? While it's still true, it's a bit confusing
Screenshot 2023-05-18 at 5 31 18 PM

src/components/QueryEditor/MetricAggregationsEditor/utils.ts
```diff
@@ -256,3 +258,9 @@ export const getChildren = (metric: MetricAggregation, metrics: MetricAggregatio
 
   return [...children, ...children.flatMap(child => getChildren(child, metrics))];
 };
+
+// TODO: Define better types for the following
+export const orderOptions = [
```
idastambuk, line 265:
```suggestion
export const orderOptions: Array<SelectableValue<string>> = [
```

src/components/QueryEditor/MetricAggregationsEditor/SettingsEditor/index.tsx
idastambuk: Removed this altogether, since afaik InlineField doesn't need an explicit arial-label

Issue #96, opened 2023-01-24
Document IAM Policy Permissions 


sarahzinger:
It would be nice to document the required minimum AWS policy permissions to help folks get up and running if they want to connect to an AWS version of open search, similar to what we have with athena and other aws datasources https://github.com/grafana/athena-datasource#authentication

PR #97, opened 2023-01-30
Docs: add access policies for OpenSearch Service and OpenSearch Serverless

kevinwcyu:
This PR documents the minimal set of permissions required to use the opensearch datasource with OpenSearch Service and OpenSearch Serverless.

Fixes: #96

Changes:
- Updated references of `Elasticsearch Service` to `OpenSearch Service`
- Add different informational sections on access policies for OpenSearch Service and OpenSearch Serverless

Notes to reviewer:
I tested these permissions the default query for the datasource. I don't know if there are other permissions required depending on the query or some other settings.

References:
- [Identity and Access Management in Amazon OpenSearch Service](https://github.com/grafana/opensearch-datasource/issues/96)
- [Overview of security in Amazon OpenSearch Serverless](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/serverless-security.html)

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,6 +288,8 @@ This is an example of a minimal policy you can use to query OpenSearch Service.
 
 ## Amazon OpenSearch Serverless
 
+> **Note:** OpenSearch Serverless support is only available in Grafana v9.4+.
+
 Access to OpenSearch Serverless data is controlled by [data access policies](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/serverless-data-access.html). These policies can be created via the Console or aws-cli.
 
 ### Data access policies for OpenSearch Serverless
--- a/README.md
+++ b/README.md
@@ -247,18 +247,72 @@ jsonData:
   pplEnabled: false
 ```
 
-## Amazon Elasticsearch Service
+## Amazon OpenSearch Service
 
-AWS users using Amazon's Elasticsearch Service can use this data source to visualize Elasticsearch data.
-If you are using an AWS Identity and Access Management (IAM) policy to control access to your Amazon Elasticsearch Service domain, then you must use AWS Signature Version 4 (AWS SigV4) to sign all requests to that domain.
+AWS users using Amazon's OpenSearch Service can use this data source to visualize OpenSearch data.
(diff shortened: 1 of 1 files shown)


sarahzinger (approved):
This looks good to me! Should we make a comment you think to say Open Search Serverless should be supported in Grafana 9.4 and above? cc @svennergr @gabor

gabor:
@sarahzinger wrote:
> Should we make a comment you think to say Open Search Serverless should be supported in Grafana 9.4 and above?

good idea! 

hmm.. though..the question is, what should we say in the required-grafana-version setting in plugin.json : https://github.com/grafana/opensearch-datasource/blob/main/src/plugin.json#L30 ?

possibilities:
- set it to grafana9.4.0... if we do that, maybe we don't need to mention this in the `README`
- keep it as it is, and just accept that on older grafanas opensearch-serverless is just not going to work.

what do you think?

kevinwcyu:
> @sarahzinger wrote:
> 
> > Should we make a comment you think to say Open Search Serverless should be supported in Grafana 9.4 and above?
> 
> good idea!
> 
> hmm.. though..the question is, what should we say in the required-grafana-version setting in plugin.json : https://github.com/grafana/opensearch-datasource/blob/main/src/plugin.json#L30 ?
> 
> possibilities:
> 
> * set it to grafana9.4.0... if we do that, maybe we don't need to mention this in the `README`
> * keep it as it is, and just accept that on older grafanas opensearch-serverless is just not going to work.
> 
> what do you think?

I'm leaning towards keeping the required grafana version as-is because the opensearch-service part should still work for older versions and people who actively want to use opensearch-serverless will have to make the decision to upgrade grafana or not.

kevinwcyu:
@gabor I'm trying to tag the docs squad to review this, but they don't show up in reviewers dropdown for me. I think they need to be added as collaborators on this repo to show up in the reviewers dropdown. Would you be able to add them please?

svennergr:
> @gabor I'm trying to tag the docs squad to review this, but they don't show up in reviewers dropdown for me. I think they need to be added as collaborators on this repo to show up in the reviewers dropdown. Would you be able to add them please?

Done

JStickler (approved):
[Docs squad]  LGTM

Issue #98, opened 2023-02-08
Error `Unknown key for a VALUE_NUMBER` when using EWMA/Holt-Winter with Moving Average


winglot:
When trying to use Moving Average aggregation with EWMA or Holt-Winters model (https://opensearch.org/docs/latest/opensearch/pipeline-agg/#moving_avg), the query fails with:
```
OpenSearch error: Unknown key for a VALUE_NUMBER in [3]: [alpha].
```

Tried this on both Grafana 8 and plugin version v1.2.4 and Grafana 9 with plugin v2.0.4, with the same error.

I have tracked down the error and it seems it is caused by using `parseInt` for `alpha`/`betta`/`gamma` parameters while OpenSearch expects a float.
(https://github.com/grafana/opensearch-datasource/blob/v2.0.5/src/components/QueryEditor/MetricAggregationsEditor/SettingsEditor/MovingAverageSettingsEditor.tsx#L44).
I think that changing the `parseInt` to `parseFloat` should fix the problem.

The Elasticsearch plugin works fine with the same query using Moving Average with EWMA as they do not use `parseInt` when getting the value for alpha/beta/gamma settings (https://github.com/grafana/grafana/blob/v9.3.6/public/app/plugins/datasource/elasticsearch/components/QueryEditor/MetricAggregationsEditor/SettingsEditor/MovingAverageSettingsEditor.tsx#L50).

PR #99, opened 2023-02-21
fix the moving_avg modes

gabor:
(fixes https://github.com/grafana/opensearch-datasource/issues/98)

when using the `moving_avg` metric, for certain modes there are coefficients (`alpha`, `beta`, `gamma`) that can be configured. the code used `parseInt` to parse them, so when floating-point values are entered (for example `0.35`), it was parsed incorrectly).

also the `MovingAverageEWMAModelSettings` type was defined incorrectly.

how to test:
1. make sure you have an opensearch datasource plugin configured ( read https://github.com/grafana/opensearch-datasource/blob/main/devenv/README.md if unsure)
1. open the opensearch datasource in explore
2. you should see a working "count" metric query
3. add another metric, choose `moving_avg`. configure it:
   - `field`: choose `count`
   - in `options`, choose `model=holt-winters`
   - set some values, for example:
       - window=5, predict=2, alpha=0.1, beta=0.2, gamma=0.3 (make sure `alpha` `beta` and `gamma` are floating-point values, not integers
4. make sure the browser-devtools are open
5. run the query
6. verify that it did not crash
7. look at the ajax request (`_msearch`)'s request-payload, it contains two lines of json, the second json should contain the correct `alpha`,`beta` and `gamma` values.
8. change the model to `exponentially weighted`, and repeat the test
9. change the model to `holt linear`, and repeat the test
10. go to dashboards, create a panel, and create the above-mentioned query there. verify it works. save the panel, then reload the page, and verify that the query is persisted correctly, meaning `alpha`, `beta` and `gamma` contain the correct values after a page reload. test all 3 models (`holt-winters`, `holt-linear`, `exponentially-weighted`)

### DIFF
--- a/src/components/QueryEditor/MetricAggregationsEditor/SettingsEditor/MovingAverageSettingsEditor.tsx
+++ b/src/components/QueryEditor/MetricAggregationsEditor/SettingsEditor/MovingAverageSettingsEditor.tsx
@@ -41,8 +41,15 @@ export const MovingAverageSettingsEditor = ({ metric }: Props) => {
         <>
           <InlineField label="Alpha">
             <Input
-              onBlur={e => dispatch(changeMetricSetting(metric, 'alpha', parseFloat(e.target.value!)))}
-              defaultValue={metric.settings?.alpha}
+              onBlur={e =>
+                dispatch(
+                  changeMetricSetting(metric, 'settings', {
+                    ...metric.settings?.settings,
+                    alpha: parseFloat(e.target.value!),
+                  })
+                )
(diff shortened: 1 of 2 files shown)


fridgepoet (approved):

kevinwcyu (approved):
lgtm. saw the values in the json payload as floats.

kevinwcyu:
Hi @gabor, can this be merged?

gabor:
hi @kevinwcyu , yes, i think it is ready to be merged.

PR #118, opened 2023-03-30
Create httpClient with grafana-plugin-sdk-go methods

### LINKED ISSUE #111: Alert fails on dashboard using opensearch datasource with TLS client auth enabled
When testing an alert created using Opensearch as datasource for a dashboard panel, we get an error that occurs while testing the alert rule that is configured:

**tsdb.HandleRequest() error rpc error: code = Unknown desc = invalid character 'U' looking for beginning of value**

image

We have configured the Opensearch datasource used in Grafana as below.

datasources:
    - name: opensearch-ds
      type: grafana-opensearch-datasource
      version: 2
      access: proxy
      orgId: 1
      url: {{ .Values.opensearch.url }}
      jsonData:
        interval: Daily
        timeField: "@timestamp"
        timeInterval: "10m"
        tlsAuth: true
        tlsAuthWithCACert: true
        database: "[opensearch-ds-]YYYY.MM.DD"
        version: "1.0.0"
        tlsSkipVerify: true
      secureJsonData:
        tlsCACert: $ENV_VAR_1
        tlsClientCert: $ENV_VAR_2
        tlsClientKey: $ENV_VAR_3

image

Please help us in getting this issue fixed at the earliest.

Grafana version: 7.5.17
Opensearch Datasource Plugin version: 1.1.2
Opensearch version: 2.4.1

**Note:** The alert does not work and it fails with same error even with the latest Grafana version 9.3.6 and the latest Opensearch datasource plugin version 2.0.5. 

Please note although that we want the fix for the older version of Grafana (v7.5.17) and Opensearch datasource plugin (v1.1.2) as mentioned above in the issue description.

fridgepoet:
* Fixes the path name of the code-coverage check
* Updates `grafana-plugin-sdk-go`
* Creates the httpClient using the `grafana-plugin-sdk-go` methods
   * This will include the TLS config which was missing before in the backend query method. Because Alerting uses the backend, this was probably creating the reported issue.

What the reporter saw was a json decoding error in the backend -- what we know is that the response body started with a `U`. In my reproduction, the full body was `Unauthorized`, probably the same for the reporter.

Fixes #111 

TODO: 
- [x] check serverless sigV4 authentication is OK


fridgepoet:
Manual check locally with serverless sigV4 authentication was OK, see internal provisioning repo for example datasource (may still be a PR)

kevinwcyu (approved):
Great investigation and work on this! Tested it and don't see the error message any more.

PR #125, opened 2023-04-05
Add 'Use time range' option, skip date type field validation

z0h3:
1) When we have many indices in elasticsearch with different names of `date` type (not universally @timestamp) field.
2) In some cases we have indices that doesn't contain any `date` type field.

We can't search data for this cases.

1) This PR allows to skip validation for `date` field, and doesn't block adding new datasource.
image

2) This PR also adds switch option `Use time range` for `raw_data` and `raw_document` search type. This option includes/excludes time range filter and sort in query request for `raw_data` and `raw_document` types only. Default value is enabled.
image

### DIFF
--- a/src/QueryBuilder.ts
+++ b/src/QueryBuilder.ts
@@ -139,10 +139,13 @@ export class QueryBuilder {
     return filterObj;
   }
 
-  documentQuery(query: any, size: number) {
+  documentQuery(query: any, size: number, useTimeRange = true) {
     query.size = size;
-    query.sort = {};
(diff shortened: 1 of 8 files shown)


sarahzinger:
@z0h3 thank you so much for your contribution! This is a great suggestion! Would you mind signing our CLA? Just follow the link in the above comment.

z0h3:
@sarahzinger Done!

sarahzinger (approved):
Looks Great! Thank you so much!

Issue #124, opened 2023-04-04
Trace Analytics: Traces Query and Lucene Query String


sarahzinger:
We have a work in progress branch right now with our work to display traces: https://github.com/grafana/opensearch-datasource/pull/122

One thing that needs further investigation is this lucene query field and it's relationship to trace queries. Some open questions:
- does typing anything in the lucene query field have any effect on trace queries right now? 
- what sorts of lucene queries might you run against all your traces?
- is there ever a situation where you might run a lucene query against a single trace?
- would it make more sense to use the lucene query field to filter down to one trace?

Issue #123, opened 2023-04-04
 Make TraceIds in Trace List View Clickable and link out to individual trace views


sarahzinger:
The goal here is to take our traces list table and transform each traceid into a link that opens a single trace view.

We have a rough version of a table of traces, and the individual trace view, but we do not connect them yet the way we do in X-Ray.

@idastambuk and I started exploring this a bit but haven't totally figured out what's missing: https://github.com/grafana/opensearch-datasource/tree/traces-query-editor-links

PR #122, opened 2023-04-04
Support for Trace Analytics

sarahzinger:
**What this PR does / why we need it**:
Adds support for [Trace Analytics](https://opensearch.org/docs/2.6/observing-your-data/trace/index/):

Screenshot 2023-04-19 at 3 14 31 PM

As part of this work we also redesigned the query editor's informational architecture. It is currently a bit confusing that queries like "raw data" and "logs" are under a metric field and we didn't want to continue that pattern with traces. In the future we hope to move out non-metric related queries from out of the metric dropdown, however to minimize changes we haven't done so yet.

**Which issue(s) this PR fixes**:
This pr contains merges from several other prs that have captured the individual issues.

PR #129, opened 2023-04-07
Trace Analytics: Add support for querying traces by lucene queries

sarahzinger:
Based off of: https://github.com/grafana/opensearch-datasource/pull/122

While working on https://github.com/grafana/opensearch-datasource/pull/122 we did not hook up the lucene query string input field. This pr adds the ability for a user to continue to filter traces results with the lucene query field. While doing so, I realized it might make more sense to combine this field and the traceid field as I'm not sure there's a use case for users wanting to filter their spans with a lucene query, nor am I confident there's a usecase to show a single traceId in the trace list view. 

While I was working on this, I kind of realized it might not make sense to store the entire lucene query object that we send to the backend within grafana. Instead we simply store that this is a traces query. 

Then in datasource.ts's query function we check if the query is a traces query and if so return back the query formatted in DSL. Then when we map responses to dataframes, we do a similar check to see if it's a trace query and whether it is a single trace query and format the response. 

In doing this I was able to delete the TracesQueryEditor and I ended up moving some of the traces files out of components and more top level as they are no longer involved in react. I apologies for the big git diff, but we could probably use another review anyway.

Fixes https://github.com/grafana/opensearch-datasource/issues/124
Fixes https://github.com/grafana/opensearch-datasource/issues/123

https://user-images.githubusercontent.com/6620164/230680542-66ae46ea-e849-4da6-8de5-21fb7950168b.mov

Also quickly added the ability to make clickable links!:

https://user-images.githubusercontent.com/6620164/230903508-53824222-1b65-48b9-8dff-be85eef2c5cf.mov


idastambuk (approved):
I really like the improvements, great job!

PR #133, opened 2023-04-12
Add tests for trace view data frames

idastambuk:

Fixes #


sarahzinger (approved):
Cool stuff Ida!

This is a bit nitpicky so please take it or leave it: my general preference when it comes to this sort of thing is to only test the top level exported function. My reasoning with that is that it might be that we someday change up something in transformTraceResponse that causes a test to fail, but that doesn't necessarily prove something is broken. What we want is to test that we can take a trace and turn it into a dataframe, and the sort of particulars of how we go about that can be changed pretty frequently but hopefully the tests that prove we can handle creating a dataframe do not change.

idastambuk:
> Cool stuff Ida!
> 
> This is a bit nitpicky so please take it or leave it: my general preference when it comes to this sort of thing is to only test the top level exported function. My reasoning with that is that it might be that we someday change up something in transformTraceResponse that causes a test to fail, but that doesn't necessarily prove something is broken. What we want is to test that we can take a trace and turn it into a dataframe, and the sort of particulars of how we go about that can be changed pretty frequently but hopefully the tests that prove we can handle creating a dataframe do not change.

Yeah, that makes sense! I initially went with that approach, but though about how it would be more tricky to debug failing tests if we don't know which of the several functions was responsible for it failing. I'll think about how to have the best of both worlds and rewrite :)

sarahzinger:
Rad up to you! No pressure to rewrite more just sharing a thought! But if you feel inspired to go for it!

src/__mocks__/openSearchTraceMock.ts
```diff
@@ -0,0 +1,156 @@
+// cSpell:disable
```
sarahzinger, line 1: oh neat I didn't know you could do this I always just add weird words to the list 😆 good to know

src/traces/formatTraces.test.ts
```diff
@@ -95,4 +97,131 @@ describe('FormatTraces', () => {
...
+      expect(singleDataFrame.fields.find(field => field.name === 'operationName').values).toEqual(
+        new ArrayVector(['/driver.DriverService/FindNearest', 'SQL SELECT'])
+      );
+      expect(singleDataFrame.fields.find(field => field.name === 'startTime').values).toEqual(
+        new ArrayVector([
+          new Date('2023-04-11T11:14:31.243838Z').getTime(),
+          new Date('2023-04-11T11:14:30.717151Z').getTime(),
+        ])
+      );
+      // cSpell:disable
```
sarahzinger, line 132: maybe better to add the list of weird words rather than not use cspell? idk I guess either way haha

src/traces/formatTraces.ts
```diff
@@ -129,27 +130,27 @@ export const createTraceDataFrame = (targets, response): DataQueryResponse => {
   return { data: dataFrames, key: targets[0].refId };
 };
 
-function transformTraceResponse(spanList: OpenSearchSpan[]): TraceSpanRow[] {
+export function transformTraceResponse(spanList: OpenSearchSpan[]): TraceSpanRow[] {
```
sarahzinger, line 133: I don't think you need to export anymore right?

sarahzinger (approved):
🆒

src/traces/formatTraces.test.ts
idastambuk: I disabled this block cause these words are probably just a one-time thing specific to this test and they most likely won't be repeated anywhere else

Issue #137, opened 2023-04-13
Trace Analytics: Hide Alias


sarahzinger:
While reviewing: https://github.com/grafana/opensearch-datasource/pull/122 

Let's hide Alias when traces is selected, I don't think it currently does anything and I'm not sure it makes sense in this context.
![Screenshot 2023-04-13 at 10 34 16 AM](https://user-images.githubusercontent.com/6620164/231793757-7e8434ac-6bc7-4df2-a5df-12958116b43a.png)
![Screenshot 2023-04-13 at 10 34 32 AM](https://user-images.githubusercontent.com/6620164/231793761-496aecfb-e88e-4d26-bcdb-7bc7683c2f4e.png)

PR #139, opened 2023-04-17
137 trace analytics hide alias

idastambuk:

Fixes https://github.com/grafana/opensearch-datasource/issues/137


sarahzinger (approved):
Looks good! Might be nice to add a test if we can and also link the original issue to the pr!

PR #148, opened 2023-04-19
Update backend dependencies

fridgepoet:
Contributes to https://github.com/grafana/oss-plugin-partnerships/issues/170


sarahzinger:
Manually tested and it seemed to work for me! Although I'm seeing a build error in the ci about the ioutil package being deprecated. Why would it build for me with mage and not in CI? Maybe I'm on a wrong version of go or something?

fridgepoet:
hey @sarahzinger thanks for having a look. this is not strictly required for https://github.com/grafana/oss-plugin-partnerships/issues/170 so no rush or anything

The binary will be built (mage or go build I guess) whether that is deprecated or not. The tests will also pass. What we're seeing is a linter error and if you're like me you don't run the linter locally.
I'll make a change for this and push it up later!

iwysiu (approved):
lgtm!

PR #151, opened 2023-04-20
Prepare for 2.4.0 release

sarahzinger:
Let me know if I missed anything!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,6 @@ All notable changes to this project will be documented in this file.
 ### Bug Fixes:
 
 - Update Backend Dependencies [#148](https://github.com/grafana/opensearch-datasource/pull/148), @fridgepoet
-- Cspell bug fix [#146](https://github.com/grafana/opensearch-datasource/pull/146), @iwysiu
 - Fix view of nested array field in table column [#128](https://github.com/grafana/opensearch-datasource/pull/128), [@z0h3](https://github.com/z0h3)
 
 ## v2.3.0


CHANGELOG.md
```diff
@@ -2,9 +2,22 @@
...
+
+### Bug Fixes:
+
+- Update Backend Dependencies [#148](https://github.com/grafana/opensearch-datasource/pull/148), @fridgepoet
+- Cspell bug fix [#146](https://github.com/grafana/opensearch-datasource/pull/146), @iwysiu
+- Fix view of nested array field in table column [#128](https://github.com/grafana/opensearch-datasource/pull/128), [@z0h3](https://github.com/z0h3)
+
 ## v2.3.0
-* Add 'Use time range' option, skip date type field validation by @z0h3 in https://github.com/grafana/opensearch-datasource/pull/125
-* Create httpClient with grafana-plugin-sdk-go methods by @fridgepoet in https://github.com/grafana/opensearch-datasource/pull/118
```
fridgepoet, line 7: This formatting with asterisks is coming from the output of generated release notes. https://github.com/grafana/opensearch-datasource/blob/6d83bfa53b57d6141d4bd6052fde0c1196375453/package.json#L13 If we add the `changelog` label to our PRs, increment the version, then the output will include the change. We copy and paste it here. I forgot about this but it would be great if we can start using it. 

Would it be bothersome to have a mix of asterisks or dashes in the markdown? I totally understand if that would be irritating to some-- in which case I propose we just change them all at once to asterisks (not necessarily in this PR!)

CHANGELOG.md
```diff
@@ -2,9 +2,22 @@
...
+## v2.4.0
+
+### Features and Enhancements:
+
+- Support for Trace Analytics [#122](https://github.com/grafana/opensearch-datasource/pull/122), @idastambuk, @katebrenner, @iwysiu, @sarahzinger
+
+### Bug Fixes:
+
+- Update Backend Dependencies [#148](https://github.com/grafana/opensearch-datasource/pull/148), @fridgepoet
+- Cspell bug fix [#146](https://github.com/grafana/opensearch-datasource/pull/146), @iwysiu
```
fridgepoet, line 14: In my opinion, adding a username to the spellcheck omission doesn't need to be part of the changelog as we should keep this fairly concise. Not sure about the backend dependencies above either.

fridgepoet (approved):

CHANGELOG.md
sarahzinger: Ooh very neat I didn't see this script! I just tried running it locally and it missed one of the prs? I don't know if I just didn't run it correctly or something.

For the asterisk/dash thing, I definitely don't care if we have a mix! It's something in my IDE that is auto changing them all, I assume it's something with the prettier config? I think rather than fussing with it, I'll let it go for now but maybe there's a hackday project here to streamline this whole flow better!

fridgepoet: Cool that's fine by me!!

fridgepoet: > I just tried running it locally and it missed one of the prs

I think we need to add how to use it somewhere in contributing. I think Adam/Andreas already wrote that up somewhere, will try to find it...

PR #152, opened 2023-04-24
Multiple queriesv3

idastambuk:

Fixes #

### DIFF
--- a/src/datasource.ts
+++ b/src/datasource.ts
@@ -522,27 +522,28 @@ export class OpenSearchDatasource extends DataSourceApi<OpenSearchQuery, OpenSea
 
     return from(this.post(this.getMultiSearchUrl(), payload)).pipe(
       map((res: any) => {
-        const er = new OpenSearchResponse(targets, res);
-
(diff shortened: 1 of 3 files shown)


sarahzinger (commented):
seems neat code wise, but taking a step back, what is the problem we're trying to solve here? Like what is the intended user experience we are fixing? 

I just made multiple queries for example and got back an error: 
Screenshot 2023-04-24 at 11 27 25 AM

but honestly I'm not terribly confident I understand what the experience should be for multiple queries you know?

idastambuk:
Closing this after thinking about this line: 
`const er = new OpenSearchResponse([target], res);`
and realizing it would probably be not so simple to extract responses that are in response.aggregations and response.hits that correspond to the respective queries. That is, without refactoring [OpenSearchResponse.ts](https://github.com/grafana/opensearch-datasource/blob/dbbb683d30dedd5be2e5174f3378bf3374984c94/src/OpenSearchResponse.ts#L535) and probably making it more unreadable. 
I think [this PR](https://github.com/grafana/opensearch-datasource/pull/141) is a less impactful solution for now

Issue #131, opened 2023-04-11
Create resource handler to get version


iwysiu:
We should create a resource handler that returns the version.

Running a get request to the url (with no query parameters) will return a version object:
'''
{
  "name" : "15ffcf528ee3",
  "cluster_name" : "docker-cluster",
  "cluster_uuid" : "1vbTN7ROS4Sw-8mHbTAIVQ",
  "version" : {
    "distribution" : "opensearch",
    "number" : "2.6.0",
    "build_type" : "tar",
    "build_hash" : "7203a5af21a8a009aece1474446b437a3c674db6",
    "build_date" : "2023-02-24T18:58:37.352296474Z",
    "build_snapshot" : false,
    "lucene_version" : "9.5.0",
    "minimum_wire_compatibility_version" : "7.10.0",
    "minimum_index_compatibility_version" : "7.0.0"
  },
  "tagline" : "The OpenSearch Project: https://opensearch.org/"
}
'''

PR #153, opened 2023-04-24
OpenSearch: Add config button to detect opensearch version

### LINKED ISSUE #132: Add button to auto-detect version to config editor
Change the ConfigEditor to have a button to detect the version. Instead of having the version selected from a dropdown, we have a button to detect the version and set it. When the button is pushed, we can save the datasource, fetch the OpenSearch version, and re-save the datasource with that information. Note that the datasource needs to be saved first, so that the backend has the correct http address and credentials to send the request. We should still display the version in the editor
We should also change testDatasource to fail for ElasticSearch v7.10+ and v8.x and point the user toward the ElasticSearch datasource, and to default to a null version instead of OpenSearch 1.0.x, to make it clearer when it hasn't been set.

iwysiu:
Currently, opensearch uses a drop down menu in the config form to get the version. This is inconvenient for users and requires us to maintain the list of OpenSearch versions (which we aren't really doing). This commit exchanges the menu for a button that saves the config (to make sure it connects properly) and detects the version. It's behind the `opensearchDetectVersion` feature flag.

Before:
Screenshot 2023-03-29 at 10 12 59 AM

The new ui:
Screenshot 2023-04-24 at 5 41 54 PM

When the call to get the version fails:
Screenshot 2023-04-24 at 5 41 36 PM

It was also changed to default to a null version (so the user has to set it to successfully configure the datasource) and for test to error when the elasticsearch version is >= 7.11.0. I can remove those changes if we think they're unnecessary.

Error for a too high elasticsearch version:
Screenshot 2023-04-25 at 1 47 47 PM

Fixes #132 #131 

If you want to test this with elasticsearch, you can run `make devenv sources=elastic` in the main grafana repo to connect to `http://localhost:9200/` or `make devenv sources=elastic elastic_version=7.10.0` to pick a specific version. (Run `make devenv-down` to take them down after)


src/__mocks__/openSearchDatasource.ts
```diff
@@ -0,0 +1,50 @@
+import { DataSourceInstanceSettings, DataSourcePluginMeta, PluginMetaInfo, PluginType } from '@grafana/data';
```
iwysiu, line 1: Maybe excessive? But I figured we might get use out of it in the future.

src/__mocks__/DefaultConfigOptions.ts
```diff
@@ -2,7 +2,9 @@ import { DataSourceSettings } from '@grafana/data';
 import { Flavor, OpenSearchOptions } from '../types';
 import { createDatasourceSettings } from '../dependencies/mocks';
 
-export function createDefaultConfigOptions(): DataSourceSettings<OpenSearchOptions> {
```
iwysiu, line 5: Moved this file for neatness

src/datasource.ts
```diff
@@ -122,10 +122,9 @@ export class OpenSearchDatasource extends DataSourceApi<OpenSearchQuery, OpenSea
         if (err.data) {
           const message = err.data.error?.reason ?? err.data.message ?? 'Unknown error';
 
-          throw {
-            message: 'OpenSearch error: ' + message,
-            error: err.data.error,
-          };
+          let newErr = new Error('OpenSearch error: ' + message);
```
iwysiu, line 125: I needed to change this to an `Error` type to catch it properly in `OpenSearchDetails.tsx`

src/configuration/OpenSearchDetails.tsx
```diff
@@ -18,14 +20,72 @@ const indexPatternTypes = [
...
+    }
+  };
+
+  let versionString = '';
+  if (value.jsonData.flavor && value.jsonData.version) {
+    versionString = `${AVAILABLE_FLAVORS.find(f => f.value === value.jsonData.flavor)?.label ||
+      value.jsonData.flavor} ${value.jsonData.version}`;
+  }
+
+  const getServerlessSettings = (event: React.SyntheticEvent<HTMLInputElement, Event>) => {
```
iwysiu, line 63: All of the query code assumes that a version and flavor are set, so rather than refactor that in this commit, I set the flavor and version when serverless is toggled. (I checked and you cannot get a version from serverless). I figured just setting it to the latest one was ok, as we'll need to go make other code changes anyway if a new latest version introduces breaking changes.

src/configuration/ConfigEditor.tsx
```diff
@@ -24,6 +25,36 @@ export const ConfigEditor = (props: Props) => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
+  const [saved, setSaved] = useState(!!options.version && options.version > 1);
+  const datasource = useDatasource(props);
+
+  useEffect(() => {
+    setSaved(false);
+  }, [
+    options.url,
```
iwysiu, line 34: This is probably excessive, but since the user has to click the button to start saving it, I figured it was better to err on the side of re-saving too often.

src/configuration/OpenSearchDetails.test.tsx
```diff
@@ -143,10 +171,15 @@ describe('OpenSearchDetails', () => {
     const onChangeMock = jest.fn();
 
     const defaultConfig = createDefaultConfigOptions();
+    const opensearchDetectVersionValue = config.featureToggles.opensearchDetectVersion;
```
sarahzinger, line 174: might make sense to take this out of the describe, it's very unlikely but I guess I could see a world where we edit config.featureToggles.opesearchDetectVersion directly in a previous test up above and accidentally never reset it or something?

src/configuration/OpenSearchDetails.tsx
sarahzinger: Do we need a version/flavor if it's serverless? Aren't all serverless the same flavor/version?

src/datasource.ts
```diff
@@ -355,6 +353,23 @@ export class OpenSearchDatasource extends DataSourceApi<OpenSearchQuery, OpenSea
...
+    if (this.flavor === Flavor.Elasticsearch && gte(this.version, '7.11.0')) {
+      return Promise.resolve({
+        status: 'error',
+        message:
+          'ElasticSearch version ' +
+          this.version +
+          ` is not supported by the OpenSearch plugin. Use the ElasticSearch plugin.`,
+      });
+    }
+    //https://grafana.com/docs/grafana/latest/datasources/elasticsearch/
```
sarahzinger, line 372: nice addition! Maybe we can share a link somehow?

sarahzinger (commented):
Looking cool!

I don't love the concept of having save and get version in the same button, but I also get that in this case it might make sense because you can't detect the version without saving your auth settings.

Another question that I'm only just thinking through now, is there a reason we have to surface this to users at all? What if testDatasource was responsible for getting/setting the opensearch version and we didn't show a field for this at all?

src/datasource.ts
iwysiu: The commented out link was from when I was trying to add the link. I couldn't get it to render properly 😞, whatever calls `testDatasource` displayed the whole href block.

iwysiu (commented):
> Another question that I'm only just thinking through now, is there a reason we have to surface this to users at all? What if testDatasource was responsible for getting/setting the opensearch version and we didn't show a field for this at all?

That was more or less proposal 1! I think we decided not to do that so that the users would be aware that the version has to be reset if they update their OpenSearch instance.

src/configuration/OpenSearchDetails.tsx
```diff
@@ -18,14 +20,73 @@ const indexPatternTypes = [
...
 export const OpenSearchDetails = (props: Props) => {
-  const { value, onChange } = props;
+  const { value, onChange, saveOptions, datasource } = props;
+  const [versionErr, setVersionErr] = useState<string>('');
+
+  const setVersion = async () => {
+    setVersionErr('');
+    await saveOptions();
+    try {
+      const version = await datasource.getOpenSearchVersion();
```
kevinwcyu, line 34: can `getOpenSearchVersion` check for supported versions and throw an error if it's >=7.11.0? right now it can successfully set a version and it appears to be ok, but i don't see an error until i save and test the datasource.

kevinwcyu (approved):
nice, tested out with es and os and works for me. just a comment about providing an error message earlier about the unsupported version when the save and get version button is pressed.

src/datasource.ts
```diff
@@ -630,6 +644,15 @@ export class OpenSearchDatasource extends DataSourceApi<OpenSearchQuery, OpenSea
     return JSON.stringify(queryObj);
   }
 
+  async getOpenSearchVersion(): Promise<{ flavor: Flavor; version: string }> {
```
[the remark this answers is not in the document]
iwysiu: done!

src/configuration/OpenSearchDetails.tsx
iwysiu: done!

iwysiu (commented):
Made Kevin's changes. Still could not figure out how to get the Alert to render with a link 😞

src/configuration/OpenSearchDetails.test.tsx
```diff
@@ -1,21 +1,45 @@
...
 import { Flavor, OpenSearchOptions } from 'types';
 import { last } from 'lodash';
 import { DataSourceSettings } from '@grafana/data';
+import { config } from '@grafana/runtime';
+import { render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { setupMockedDataSource } from '__mocks__/OpenSearchDatasource';
 const { Select, Switch } = LegacyForms;
 
+const opensearchDetectVersionValue = config.featureToggles.opensearchDetectVersion;
```
sarahzinger, line 15: yeah this is what I meant!

src/configuration/OpenSearchDetails.test.tsx
```diff
@@ -1,21 +1,45 @@
...
+
+afterAll(() => {
```
sarahzinger, line 17: does this work when it's outside of the describe? it might have to be inside? Please go ahead and tell me if I'm wrong tho haha

src/configuration/OpenSearchDetails.tsx
sarahzinger: I tested this with a serverless datasource and the flow as a bit weird: 

https://user-images.githubusercontent.com/6620164/235996739-a9d00519-291c-4bd2-84c7-2001e46e5dc7.mov

what if when we get/set version we check if it's serverless based on the url and then set the toggle? 

I think we should be able to dynamically determine it based on the url, I think for example most of the endpoints are formatted like `https://{someidorsomething}.{region}.aoss.amazonaws.com`

I wonder if we could do a regex for `aoss.amazonaws.com` and if it's there we set the version to serverless or something?

sarahzinger (approved):
Cool stuff! 

I was playing around with this and one thing I could see happening is users click Save and Get Version misread this as Get and Save and think that their version is saved as well? I do sort of wonder if we should just call it Get Version after all? Or alternatively we could save again after the version is set? Idk what do you think? Am I overthinking this? It's a tricky ui question here
 
https://user-images.githubusercontent.com/6620164/236001505-5ae14381-2d8d-4e30-a8e3-a0e22f3051fe.mov

Also had an idea for optimizing the serverless flow. But also these suggestions feel like things we could explore in future issues/prs if we want to merge as is I'm down!

iwysiu:
Edited after further discussion to:
- save after getting the version
- have an info box explicitly call out what happens when upgrading
- moves serverless toggle above the version field, so hopefully users who need that toggle it first

Screenshot 2023-05-04 at 3 02 46 PM

kevinwcyu (approved):

src/datasource.test.ts
```diff
@@ -1179,6 +1197,42 @@ describe('OpenSearchDatasource', function(this: any) {
...
+      const version = await ctx.ds.getOpenSearchVersion();
+      expect(version.flavor).toBe(Flavor.Elasticsearch);
+      expect(version.version).toBe('7.6.0');
+    });
+
+    it('should error for invalid version', async () => {
+      datasourceRequestMock.mockImplementation(() => {
+        return Promise.resolve({ data: { version: { number: '7.11.1' } } });
+      });
+      //'ElasticSearch version 7.11.1 is not supported by the OpenSearch plugin. Use the ElasticSearch plugin.'
```
katebrenner, line 1230: is this comment left over from development?

katebrenner (approved):

PR #166, opened 2023-05-09
Apply default version to a serverless config

iwysiu:
An edge case that I didn't catch while developing. It can be reproduced by using the aws provisioner app to provision a serverless OpenSearch version (we have one in `US East (Ohio)`) and opening and trying to save the config. Before this fix, we would get the `No version set` error, but this applys a version when the config editor is open (which is what it did before the detecting version change). We don't really have any tests for this code path at the moment, but I can whip together a `coerceOptions` test if we think we need one.

### DIFF
--- a/src/configuration/utils.ts
+++ b/src/configuration/utils.ts
@@ -9,7 +9,7 @@ export const coerceOptions = (
 ): DataSourceSettings<OpenSearchOptions, {}> => {
   let version = valid(options.jsonData.version);
   let flavor = options.jsonData.flavor;
-  if (!config.featureToggles.opensearchDetectVersion) {
+  if (!config.featureToggles.opensearchDetectVersion || options.jsonData.serverless) {
     flavor = flavor || Flavor.OpenSearch;
     version =
       version ||


sarahzinger (commented):
I'm thinking in the future we will remove the feature toggle but still need to set a default option for serverless but in that case we won't need to do all of this right since the version/flavor are more obvious? Maybe we should write the code for serverless as a separate if conditional that we can keep as is after we remove the feature flag, wdyt?

iwysiu:
> I'm thinking in the future we will remove the feature toggle but still need to set a default option for serverless but in that case we won't need to do all of this right since the version/flavor are more obvious? Maybe we should write the code for serverless as a separate if conditional that we can keep as is after we remove the feature flag, wdyt?

I think this is what you meant? In the long term, we should probably refactor the code to handle serverless cases without a dummy version, but I figured that could wait until we decide whatever refactoring we want to do to opensearch.

src/configuration/utils.ts
```diff
@@ -16,6 +16,10 @@ export const coerceOptions = (
       AVAILABLE_VERSIONS.find(v => v.value.flavor === flavor)?.value.version ||
       AVAILABLE_VERSIONS[AVAILABLE_VERSIONS.length - 1].value.version;
   }
+  if (options.jsonData.serverless) {
+    flavor = flavor || Flavor.OpenSearch;
+    version = version || '1.0.0';
+  }
```
sarahzinger, line 22: so why is this an Or case? I thought in serverless was always the same?

sarahzinger (approved):

src/configuration/utils.ts
iwysiu: done, it was in the old code and I didn't think about it

idastambuk (approved):

Issue #117, opened 2023-03-30
Tests are disabled and out of date in response_parser_test.go


fridgepoet:
**What happened**:
`response_parser_test.go_` has an underscore at the end of the file name. https://github.com/grafana/opensearch-datasource/blob/main/pkg/opensearch/response_parser_test.go_

Not sure if this was intentional; it's been there since the first commit of this repo.

The consequence is that no test in that file is run under `go test`. 

* Rename it without the `_` at the end. 
* All of the tests need to be verified and assertions need to be updated.

fridgepoet:
This is what the test was copied from originally, can probably use it to update some of the tests:
https://github.com/grafana/grafana/blob/main/pkg/tsdb/elasticsearch/response_parser_test.go

fridgepoet:
Closing this now but a few more tests need to be re-enabled https://github.com/grafana/opensearch-datasource/issues/175

PR #174, opened 2023-05-17
Enable response_parser_test.go

fridgepoet:
* enables the tests that were in response_parser_test.go_
* updates the test assertions from using Points to now Frames and Fields
* converts the assertions to use the testify library
* renames a package from `es` to `client`

Discovered that some table to data frame conversion has not yet been done in this plugin. The tests that depend on that are now commented and an issue was created https://github.com/grafana/opensearch-datasource/issues/175.

Contributes to https://github.com/grafana/opensearch-datasource/issues/117

**Other information**

Something I found interesting was that time in the "Points" used to be stored as float64 epoch time, so a "1000" time was the same as time.UnixMilli(1000).UTC() which is the same as time.Date(1970, time.January, 1, 0, 0, 1, 0, time.UTC). These are all typed now in the time-typed Fields.


pkg/opensearch/response_parser_test.go
```diff
@@ -0,0 +1,1030 @@
...
+		assert.Nil(t, err)
+		assert.Len(t, result.Responses, 1)
+
+		queryRes := result.Responses["A"]
+		assert.NotNil(t, queryRes)
+		assert.Len(t, queryRes.Frames, 6)
+
+		seriesOne := queryRes.Frames[0]
+		assert.Equal(t, "server1 Max", seriesOne.Name)
+		assert.Len(t, seriesOne.Fields, 2)
```
iwysiu, line 389: nit: I think that the intent of the original was checking the number of data points rather than the number of fields on this, so maybe a line like `assert.Len(t, seriesOne.Fields[0], 1)` should be used?

pkg/opensearch/response_parser_test.go
```diff
@@ -0,0 +1,1030 @@
...
+		seriesThree := queryRes.Frames[2]
+		assert.Equal(t, "0 Count and {{not_exist}} 0", seriesThree.Name)
+		assert.Len(t, seriesThree.Fields, 2)
+		assert.Equal(t, time.Date(1970, time.January, 1, 0, 0, 1, 0, time.UTC), *seriesThree.Fields[0].At(0).(*time.Time))
+		assert.Equal(t, time.Date(1970, time.January, 1, 0, 0, 2, 0, time.UTC), *seriesThree.Fields[0].At(1).(*time.Time))
+		assert.EqualValues(t, 2, *seriesThree.Fields[1].At(0).(*float64))
+		assert.EqualValues(t, 8, *seriesThree.Fields[1].At(1).(*float64))
+	})
+
+	// TODO: this test will require some conversion of tables to data frames, original work in Elasticsearch https://github.com/grafana/grafana/pull/34710
```
iwysiu, line 504: nit: It may make sense to also have the number of the opensearch issue here

iwysiu (approved):
Besides my note on what I think `So(seriesOne.Points, ShouldHaveLength, 2)` from the original is checking (I think it's checking the number of data points while you're checking the number of fields per data point) looks good!

pkg/opensearch/response_parser_test.go
fridgepoet: Nah, Fields is really at the same level that Points was and Points's length was all they were checking. 

Something that happened was organization was a little bit switched between Points and Fields.
For example Points[0][0] was a float64 and Points[0][1] was a time.
Now in Fields that would be: Fields[0][0] is time and Fields [1][0] is float64. So in the case where you only are describing one "data point", you would have had a Points of length 1, but now you would have a Fields of length 2.

However before they never checked the second dimension length. (They would have had a panic if they were then accessing a second dimension that didn't exist like Points[0][<doesn't exist>].) 
So I changed the following:
* changed any length assertions before accessing indexes to "require" so that the test doesn't proceed and print panics
* added the length requirement assertions to the second dimensions

pkg/opensearch/response_parser_test.go
fridgepoet: done

sarahzinger (approved):
🚀

PR #180, opened 2023-05-22
Backend: Set field.Config.DisplayNameFromDS instead of frame.name

fridgepoet:
**What this PR does / why we need it**:

The time series legend when displayed from a backend query (alerting, expressions) was duplicating.

This PR applies a [change from Elasticsearch](https://github.com/grafana/grafana/pull/60169) to generate field names for the value field of time series frames. 

It applies the fixes from [another PR in Elasticsearch ](https://github.com/grafana/grafana/pull/66039) to move from putting the frame name in `frame.name` to `field.Config.DisplayNameFromDS` and getting the metric name from `metric.Field` instead of `metric.ID`. 

The screenshot illustrates the same label for a backend flow and frontend flow. 
Screenshot 2023-05-22 at 18 48 19

Thank you @idastambuk for pointing out where to set the legend.

**Which issue(s) this PR fixes**:

Fixes #178

PR #182, opened 2023-05-25
Prepare v2.6.0

fridgepoet:

* Ability to select order (Desc/Asc) for "raw data" metrics aggregations in https://github.com/grafana/opensearch-datasource/pull/88
* Backend: Set field.Config.DisplayNameFromDS instead of frame.name in https://github.com/grafana/opensearch-datasource/pull/180

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "grafana-opensearch-datasource",
-  "version": "2.5.1",
+  "version": "2.5.2",
   "description": "",
   "scripts": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "grafana-opensearch-datasource",
-  "version": "2.5.1",
+  "version": "2.5.2",
```
sarahzinger, line 3: I think since this includes some new functionality, it is 2.6, yeah?

sarahzinger (approved):

Issue #196, opened 2023-06-08
Backend: Implement Lucene raw_data queries in the backend


idastambuk:
Note: Blocked by https://github.com/grafana/opensearch-datasource/issues/196

Currently, Lucene raw_data queries are processed as Lucene logs (https://github.com/grafana/opensearch-datasource/blob/ace67550f9591b0298488aac2044be96b30c89e4/src/OpenSearchResponse.ts#L447) (frontend). This flow needs to be compared and migrated to the backend. 
On the backend, raw_data is processed [in the backend in getTimeSeries but is incomplete](https://github.com/grafana/opensearch-datasource/blob/f24503df529812e054b9b2619be29aea9c381660/pkg/opensearch/response_parser.go#L45). It will need to be separated: See [this ElasticSearch refactor PR](https://github.com/grafana/grafana/pull/63208) and https://github.com/grafana/grafana/pull/59741

Some general guidelines are:

* Create visualization to compare between frontend and backend flows, preferably in a shared Cloud dashboard e.g. clouddatasources.grafana.net 
* Unit tests in the frontend should be copied or updated in the backend to assert the same behavior and add unit tests when there are none.
    * Check Elasticsearch repo for related tests – helpful when there are no OpenSearch tests in the frontend or backend
* We should match OpenSearch's backend behavior to the current frontend behavior, but be aware it may not always make sense. Elasticsearch's current behavior is also another resource for how OpenSearch should behave.
    * Let's be pragmatic about this – our foremost goal is the migration of OpenSearch's frontend behavior. We can aim for some Elasticsearch feature parity, but only _within reason_.
* Often there is a PR in the Elasticsearch backend data source migration [https://github.com/grafana/grafana/issues/54011](https://github.com/grafana/grafana/issues/54011) which is 75% of what OpenSearch will need
* Migrate or implement: i.e. make any changes to code in backend
* Use the dashboard created above to compare visualization results (data frame comparison)
    * Can also inspect data frame on a more detailed basis

PR #203, opened 2023-06-15
Add raw_data query support to backend

fridgepoet:
This recreates the raw_data query processing that exists in the frontend in the backend. It redirects any queries with only raw_data to the backend.
This is based on equivalent work in Elasticsearch (https://github.com/grafana/grafana/pull/63208 and https://github.com/grafana/grafana/pull/59741) and on the existing query logic in OpenSearch's frontend.

1. For each hit in the `hits` response, we create an object:
   * with entries for each object in `_source` -- these are flattened to a maximum depth of 10
   * with entries `_id`, `_type`, `_index`
   * with an entry for timestamp from `fields` if possible and falling back to the timestamp `_source` if possible
2. Each entry becomes a data frame field with the values associated with each hit
   * To detect the data frame type, we look at first non-null value and set type based on it.
3. The fields are then sorted alphabetically, except for the time field which is specially placed at the beginning.

For testing, one way is to use the sample data and create a query > Metric > Raw Data.
Compare a version without these changes, e.g. v2.6.2. 
In our squad, check the data source set up here https://clouddatasources.grafana.net/goto/BTjFTmuVR?orgId=1 under the Lucene Raw Data section.
This should be running the frontend logic.

Build locally this PR and make the same query.
The same data results should be observed. There is a difference because the data frame types are now interpreted (previously values were all strings), so some default format options round the numbers.

Here are some examples of a visual comparison: 
Black is the frontend flow
Screenshot 2023-06-15 at 16 31 02
White is the locally-built backend flow
Screenshot 2023-06-15 at 16 31 55

Fixes https://github.com/grafana/opensearch-datasource/issues/196

When possible, tests were pulled from the Elasticsearch PR and adapted for our code.
We also included this change https://github.com/grafana/grafana/pull/67767 as we ran into the issue. We'll likely include it for other types of queries as well.

If there are too many changes, let me know if I should try to break it up. I'm also happy to go over it together.


pkg/opensearch/response_parser.go
```diff
@@ -74,19 +77,206 @@ func (rp *responseParser) getTimeSeries() (*backend.QueryDataResponse, error) {
...
 }
 
+func processRawDataResponse(res *es.SearchResponse, timeField string, queryRes backend.DataResponse) backend.DataResponse {
+	propNames := make(map[string]bool)
+	docs := make([]map[string]interface{}, len(res.Hits.Hits))
+
+	for hitIdx, hit := range res.Hits.Hits {
+		var flattenedSource map[string]interface{}
+		if hit["_source"] != nil {
+			flattenedSource = flatten(hit["_source"].(map[string]interface{}), 10)
```
iwysiu, line 106: nit: should comment arbitrary consts (maybe put at the top named `maxFlattenDepth`?)

pkg/opensearch/response_parser.go
```diff
@@ -74,19 +77,206 @@ func (rp *responseParser) getTimeSeries() (*backend.QueryDataResponse, error) {
...
+		if timeString, ok := source[timeField].(string); ok {
+			return timeString, true
+		}
+	}
+
+	return "", false
+}
+
+func flatten(target map[string]interface{}, maxDepth int) map[string]interface{} {
+	output := make(map[string]interface{})
```
iwysiu, line 175: Is there a reason this is a recursive function in this version? I think iterative ones usually end up being easier to read.

pkg/opensearch/response_parser.go
```diff
@@ -74,19 +77,206 @@ func (rp *responseParser) getTimeSeries() (*backend.QueryDataResponse, error) {
...
+	frames := data.Frames{}
+	frame := data.NewFrame("", fields...)
+	frames = append(frames, frame)
+
+	queryRes.Frames = frames
+	return queryRes
+}
+
+func getTimestamp(hit, source map[string]interface{}, timeField string) *time.Time {
+	timeString, ok := lookForFieldsInHit(hit, timeField)
```
iwysiu, line 131: The logic here may benefit from a comment, I had to read carefully to figure out what was actually happening here

pkg/opensearch/response_parser.go
```diff
@@ -74,19 +77,206 @@ func (rp *responseParser) getTimeSeries() (*backend.QueryDataResponse, error) {
...
+		v, ok := value.(map[string]interface{})
+		if ok && len(v) > 0 && currentDepth <= maxDepth {
+			step(nextDepth, maxDepth, v, newKey, output)
+		} else {
+			output[newKey] = value
+		}
+	}
+}
+
+func processDocsToDataFrameFields(docs []map[string]interface{}, propNames map[string]bool, timeField string) []*data.Field {
```
iwysiu, line 194: does timefield get used?

iwysiu (commented):
A couple questions/nits, but nice work!

idastambuk:
> We'll likely include it for other types of queries as well.

We should probably add this in the description of either the epic ticket or the tasks in it

fridgepoet:
Thanks @iwysiu and @idastambuk for having a look! I think I've addressed the feedback so far. Let me know if you have any other suggestions.

pkg/opensearch/response_parser.go
fridgepoet: Did you mean in this version vs Elasticsearch? Theirs is recursive.

pkg/opensearch/client/search_request.go
```diff
@@ -101,6 +101,10 @@ func (b *SearchRequestBuilder) AddDocValueField(field string) *SearchRequestBuil
 	return b
 }
 
+func (b *SearchRequestBuilder) AddTimeFieldWithStandardizedFormat(timeField string) {
```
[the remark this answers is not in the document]
iwysiu: My fault for not being clear: I meant the comment for the function. (in general exported functions should be documented). Though the other one is useful too!

pkg/opensearch/response_parser.go
iwysiu: For some reason I thought it was iterative. Thanks for updating it!

iwysiu (approved):
Nice work! One last little comment nit that was just me being unclear.

Was there a reason you needed to force push? If you can, once people have started reviewing a pr you should avoid force pushing, it messes up the links for comments. (If you need to fix merge conflicts you can do a rebase force push after its been approved)

pkg/opensearch/client/search_request.go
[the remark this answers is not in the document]
fridgepoet: Done!

pkg/opensearch/response_parser.go
fridgepoet: heya @iwysiu I think the recursive version, when refactored a bit, is much clearer. Here's what it looks like: https://github.com/grafana/grafana/pull/70302/files#diff-e269aab87b3357557da16d1389830e37476d46b47205d5c4c9b1d1d1d1473dc4R1044

Let me know if you prefer it!

pkg/opensearch/response_parser.go
```diff
@@ -74,19 +77,206 @@ func (rp *responseParser) getTimeSeries() (*backend.QueryDataResponse, error) {
 		queryRes := backend.DataResponse{
 			Frames: data.Frames{},
 		}
-		props := make(map[string]string)
-		err := rp.processBuckets(res.Aggregations, target, &queryRes, props, 0)
-		if err != nil {
-			return nil, err
+
+		switch {
+		case target.Metrics[0].Type == rawDataType || target.Metrics[0].Type == rawDocumentType:
```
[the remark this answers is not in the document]
fridgepoet: Thanks for catching this! I changed the code a couple of weeks ago after you pointed it out to me to no longer impact raw_document -- let me know if it's alright.

pkg/opensearch/response_parser.go
iwysiu: It looks nice! I'm still going to be annoying about commenting the arbitrary consts, lol. (Also if you're using it in 2 places it should be a variable)

iwysiu: I don't really have a preference at this point, we can do the recursive version to match elasticsearch if that makes more sense to you.

PR #211, opened 2023-07-12
Change query sort to respect sort order

fridgepoet:
Sort is an array, not just a map. 
This PR changes it to an array, which will reflect the order of priority defined in the code. 
The `map[string]interface{}` can also be more precisely typed as `map[string]map[string]string{}`. 

See https://opensearch.org/docs/latest/search-plugins/searching-data/sort/ for documentation on the array and examples.
>The sort parameter is an array, so you can specify multiple field values in the order of their priority.

Sort is only currently used in raw_data (and soon-to-be-implemented-in-the-backend raw_document) queries.

Similar to what was identified in Elasticsearch: https://github.com/grafana/grafana/blob/b39c8aec6c36a9557b147f628496af591cf1583e/pkg/tsdb/elasticsearch/client/search_request.go#L19C1-L19C1

Contributes to https://github.com/grafana/opensearch-datasource/issues/196

### DIFF
--- a/pkg/opensearch/client/search_request_test.go
+++ b/pkg/opensearch/client/search_request_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/Masterminds/semver"
 	simplejson "github.com/bitly/go-simplejson"
 	"github.com/grafana/opensearch-datasource/pkg/tsdb"
+	"github.com/stretchr/testify/assert"
 
 	. "github.com/smartystreets/goconvey/convey"
 )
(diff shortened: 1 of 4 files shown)


pkg/opensearch/client/search_request_test.go
```diff
@@ -88,9 +89,9 @@ func TestSearchRequest(t *testing.T) {
 						So(err, ShouldBeNil)
 						So(json.Get("size").MustInt(0), ShouldEqual, 200)
 
-						sort := json.GetPath("sort", timeField)
-						So(sort.Get("order").MustString(), ShouldEqual, "desc")
-						So(sort.Get("unmapped_type").MustString(), ShouldEqual, "boolean")
+						sort := json.Get("sort").MustArray()
+						So(sort, ShouldHaveLength, 1)
+						So(sort[0], ShouldResemble, map[string]interface{}{timeField: map[string]interface{}{"order": "desc", "unmapped_type": "boolean"}})
```
iwysiu, line 94: should the types here be `map[string]string` instead of `map[string]interface{}`?

iwysiu (approved):
Nice! If it's getting added to the changelog, maybe the pr title should be something like "Make query sort respect sort order" which is closer to what the users care about?

idastambuk (approved):

pkg/opensearch/client/search_request_test.go
fridgepoet: This tests the type that comes from simplejson and also uses some "ShouldResemble" qualifier. It doesn't want to be more precise than interface{}.

Since we just care about the json, I moved the test to the bottom of the file and tested the json string equality.

PR #214, opened 2023-07-13
Add raw_document query support to backend

fridgepoet:
This recreates the raw_document query processing that exists in the frontend in the backend. It redirects any queries with only raw_document to the backend. It's very similar to the recent [raw_data work](https://github.com/grafana/opensearch-datasource/pull/203).

The query building is based on its equivalent in [Elasticsearch](https://github.com/grafana/grafana/blob/b39c8aec6c36a9557b147f628496af591cf1583e/pkg/tsdb/elasticsearch/data_query.go#L358) and the original PR is [here](https://github.com/grafana/grafana/pull/59741).
The result processing is based on its equivalent in [Elasticsearch](https://github.com/grafana/grafana/blob/b39c8aec6c36a9557b147f628496af591cf1583e/pkg/tsdb/elasticsearch/response_parser.go#L210). The original PR is [here](https://github.com/grafana/grafana/pull/63932).

Fixes https://github.com/grafana/opensearch-datasource/issues/195

* As mentioned in the [Elasticsearch PR](https://github.com/grafana/grafana/pull/63932), the order of the keys in the JSON will not be the same between the previous frontend implementation and this backend PR. 
  > results from backend are alphabetically ordered - this is done automatically while JSON.Marshall. From [docs](https://pkg.go.dev/encoding/json#Marshal): ...The map keys are sorted. This shouldn't be an issue at all, as the order of fields in json is not used anywhere.

* Another difference is that the `timestamp` will be an array, not just the value, same as in the [Elasticsearch PR](https://github.com/grafana/grafana/pull/63932). 
  > In json field (only field), @timestamp is array with value when processed trough backend (compared to just value when processed trough frontend).

  Although we can match the same behavior as the frontend by supplying only the single value from the array, if this more direct result has been acceptable for Elasticsearch users, the code will be simpler.

### DIFF
--- a/pkg/opensearch/lucene_handler.go
+++ b/pkg/opensearch/lucene_handler.go
@@ -44,7 +44,8 @@ func (h *luceneHandler) processQuery(q *Query) error {
 	b := h.ms.Search(interval)
 	b.Size(0)
 	filters := b.Query().Bool().Filter()
-	filters.AddDateRangeFilter(h.client.GetTimeField(), es.DateFormatEpochMS, toMs, fromMs)
+	defaultTimeField := h.client.GetTimeField()
+	filters.AddDateRangeFilter(defaultTimeField, es.DateFormatEpochMS, toMs, fromMs)
 
 	if q.RawQuery != "" {
 		filters.AddQueryStringFilter(q.RawQuery, true)
@@ -58,10 +59,10 @@ func (h *luceneHandler) processQuery(q *Query) error {
 	}
 
 	switch {
-	case q.Metrics[0].Type == rawDocumentType, q.Metrics[0].Type == rawDataType:
-		processDocumentQuery(q, b, h.client.GetTimeField())
+	case q.Metrics[0].Type == rawDataType:
+		processRawDataQuery(q, b, defaultTimeField)
 	default:
-		processTimeSeriesQuery(q, b, fromMs, toMs)
+		processTimeSeriesQuery(q, b, fromMs, toMs, defaultTimeField)
 	}
 
 	return nil
@@ -69,7 +70,7 @@ func (h *luceneHandler) processQuery(q *Query) error {
 
 const defaultSize = 500
 
-func processDocumentQuery(q *Query, b *es.SearchRequestBuilder, defaultTimeField string) {
+func processRawDataQuery(q *Query, b *es.SearchRequestBuilder, defaultTimeField string) {
 	metric := q.Metrics[0]
 	order := metric.Settings.Get("order").MustString()
 	b.Sort(order, defaultTimeField, "boolean")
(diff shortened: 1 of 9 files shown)


pkg/opensearch/lucene_handler.go
```diff
@@ -59,8 +59,8 @@ func (h *luceneHandler) processQuery(q *Query) error {
 	}
 
 	switch {
-	case q.Metrics[0].Type == rawDataType:
-		processRawDataQuery(q, b, defaultTimeField)
+	case q.Metrics[0].Type == rawDocumentType, q.Metrics[0].Type == rawDataType:
```
iwysiu, line 62: Nit: Maybe i'm misremembering the syntax, but could you do:
```
switch q.Metrics[0].Type {
case rawDocumentType, rawDataType:
```

pkg/opensearch/response_parser.go
```diff
@@ -81,6 +81,8 @@ func (rp *responseParser) getTimeSeries(timeField string) (*backend.QueryDataRes
 		switch {
 		case target.Metrics[0].Type == rawDataType:
 			queryRes = processRawDataResponse(res, timeField, queryRes)
+		case target.Metrics[0].Type == rawDocumentType:
```
iwysiu, line 84: same as the above switch comment

pkg/opensearch/response_parser.go
```diff
@@ -131,6 +133,60 @@ func processRawDataResponse(res *es.SearchResponse, timeField string, queryRes b
 	return queryRes
 }
 
+func processRawDocumentResponse(res *es.SearchResponse, timeField, refID string, queryRes backend.DataResponse) backend.DataResponse {
+	documents := make([]map[string]interface{}, len(res.Hits.Hits))
+	for hitIdx, hit := range res.Hits.Hits {
+		doc := map[string]interface{}{
+			"_id":    hit["_id"],
+			"_type":  hit["_type"],
+			"_index": hit["_index"],
```
iwysiu, line 142: Do  need the `sort` and `highlight` fields that Elasticsearch is copying over?

pkg/opensearch/response_parser.go
```diff
@@ -131,6 +133,60 @@ func processRawDataResponse(res *es.SearchResponse, timeField string, queryRes b
...
+		}
+		value := json.RawMessage(bytes)
+		fieldVector[i] = &value
+	}
+
+	isFilterable := true
+	field := data.NewField(refID, nil, fieldVector)
+	field.Config = &data.FieldConfig{Filterable: &isFilterable}
+
+	frames := data.Frames{}
```
iwysiu, line 182: nit: is there a reason not to do this as `frames := data.Frames{data.NewFrame(refID, field)}`
 or even `queryRes.Frames = data.Frames{data.NewFrame(refID, field)}`

iwysiu (commented):
Mostly just style nits again, with one question about the hit fields

pkg/opensearch/response_parser.go
fridgepoet: no, these are not in OpenSearch

iwysiu (approved):

Issue #224, opened 2023-07-26
ad hoc filters not applied to raw_data queries


gulecroc:
**What happened**:

Use a table with raw data.

When adding a filter from ad hoc filter, the table must filter lines.

Or when adding a filter from the table (clic on button "+" to add the key/value as an ad hoc filter).

![image](https://github.com/grafana/opensearch-datasource/assets/8898737/27aa92dd-2bcf-4403-8575-4f32451c07f4)

I have 2 problems :
- when adding filter from the ad hoc filter on the top, there is no value _Edit: this has been separated [to another issue](https://github.com/grafana/opensearch-datasource/issues/227) now_
- when adding filter from the table value (clic on button "+" to add the key/value filter), the table is not filtered

**What you expected to happen**:

The table content is filtered with ad hoc filter.

**How to reproduce it (as minimally and precisely as possible)**:

Create opensearch data :

```
DELETE /grafana_table_raw_data_ad_hoc_filters

PUT grafana_table_raw_data_ad_hoc_filters

POST /grafana_table_raw_data_ad_hoc_filters/_bulk
{"index":{}}
{"name": "test0", "filter1": "value1", "@timestamp" : "2023-07-26"}
{"index":{}}
{"name": "test1", "filter1": "value1", "@timestamp" : "2023-07-26"}
{"index":{}}
{"name": "test2", "filter1": "value1", "@timestamp" : "2023-07-26"}
{"index":{}}
{"name": "test3", "filter1": "value2", "@timestamp" : "2023-07-26"}
{"index":{}}
{"name": "test4", "filter1": "value2", "@timestamp" : "2023-07-26"}
{"index":{}}
{"name": "test5", "filter1": "value2", "@timestamp" : "2023-07-26"}
{"index":{}}
{"name": "test6", "filter1": "value3", "@timestamp" : "2023-07-26"}
{"index":{}}
{"name": "test7", "filter1": "value3", "@timestamp" : "2023-07-26"}
{"index":{}}
{"name": "test8", "filter1": "value3", "@timestamp" : "2023-07-26"}
{"index":{}}
{"name": "test9", "filter1": "value4", "@timestamp" : "2023-07-26"}
```

Add grafana opensearch datasource from this index.

Create a simple dashboard with table and raw data :

![image](https://github.com/grafana/opensearch-datasource/assets/8898737/e75b003c-56ad-4e73-9ae8-40cbdb7b5ebf)

[...]

Issue #192, opened 2023-06-08
Migration to a Backend Data Source


fridgepoet:
```[tasklist]
### Before Node Graph
- [ ] https://github.com/grafana/oss-plugin-partnerships/issues/192
- [ ] https://github.com/grafana/oss-plugin-partnerships/issues/195
- [x] move non-reg dashboard to cloudatasource Cloud instance
- [ ] https://github.com/grafana/opensearch-datasource/issues/185
- [ ] https://github.com/grafana/opensearch-datasource/issues/175
- [ ] https://github.com/grafana/opensearch-datasource/issues/196
- [x] Prerequisite for Lucene work below: https://github.com/grafana/opensearch-datasource/issues/224
- [ ] https://github.com/grafana/opensearch-datasource/issues/195
- [ ] https://github.com/grafana/opensearch-datasource/issues/199
- [ ] https://github.com/grafana/opensearch-datasource/issues/248
- [ ] https://github.com/grafana/opensearch-datasource/issues/226
- [ ] https://github.com/grafana/oss-plugin-partnerships/issues/199
- [ ] https://github.com/grafana/opensearch-datasource/issues/255
- [ ] https://github.com/grafana/opensearch-datasource/issues/197
- [ ] https://github.com/grafana/opensearch-datasource/issues/200
- [ ] https://github.com/grafana/opensearch-datasource/issues/201
- [ ] https://github.com/grafana/opensearch-datasource/issues/156
- [ ] https://github.com/grafana/opensearch-datasource/issues/223
```

```[tasklist]
### After Node Graph
- [ ] https://github.com/grafana/opensearch-datasource/issues/296
- [ ] https://github.com/grafana/opensearch-datasource/issues/275
- [ ] https://github.com/grafana/opensearch-datasource/issues/290
- [ ] https://github.com/grafana/opensearch-datasource/issues/372
- [ ] https://github.com/grafana/opensearch-datasource/issues/373
- [ ] https://github.com/grafana/opensearch-datasource/issues/374
- [ ] https://github.com/grafana/opensearch-datasource/issues/379
- [x] https://github.com/grafana/opensearch-datasource/issues/161 (Investigate which queries this change may need to be applied to https://github.com/grafana/grafana/pull/67767)
- [ ] https://github.com/grafana/opensearch-datasource/issues/407
```

```[tasklist]
### Related bugs
- [ ] https://github.com/grafana/opensearch-datasource/issues/176
[...]

PR #225, opened 2023-07-27
Add ad hoc filters before sending Lucene queries to backend

fridgepoet:
For Lucene queries which were recently migrated to the backend, this fixes ad hoc filters on queries which were redirected to the backend as part of this plugin's [migration to a backend data source](https://github.com/grafana/opensearch-datasource/issues/192).

Ad hoc filters were only added [in the frontend while building the query](https://github.com/grafana/opensearch-datasource/blob/ebbfc716282a66592333a5252ae6453e356ebae7/src/datasource.ts#L648), so this code was not reached when [diverting queries to the backend](https://github.com/grafana/opensearch-datasource/blob/main/src/datasource.ts#L459C1-L484). 

Here is a recording of this PR's behavior:

https://github.com/grafana/opensearch-datasource/assets/4163034/44fd9e81-2897-4176-b322-2482abacda27

Fixes https://github.com/grafana/opensearch-datasource/issues/224

This PR should revert the behavior to the same as in v2.6.2 and before.
Ad hoc filtering will not return data for nested documents as this is not supported in the Lucene query language. 

Thank you @idastambuk for all your help on this one!

### DIFF
--- a/src/datasource.test.ts
+++ b/src/datasource.test.ts
@@ -1291,15 +1291,6 @@ describe('OpenSearchDatasource', function(this: any) {
     });
 
     it('does not send query including types not yet migrated to backend', () => {
-      // TODO: Ask Ida if we need this part? still compiles below
-      // datasourceRequestMock.mockImplementation(options => {
-      //   return Promise.resolve({
-      //     data: {
-      //       responses: [],
-      //     },
-      //   });
-      // });
-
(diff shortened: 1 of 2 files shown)


idastambuk (approved):
Looks good! 
If you already havent done it, maybe it would be good to add a task to the refactor epic to use the new adhoc processing for FE queries too?

Issue #199, opened 2023-06-12
Backend: Implement Lucene Logs queries in the backend


idastambuk:
Note: Blocked by https://github.com/grafana/opensearch-datasource/issues/224

Currently, there is no separate flow to process Lucene Logs on the backend. They are processed as time series in processAggregationDocs.
On the FE logs are processed in processResponseToDataFrames(). 
The output should be compared and separated out (from raw_data or in general) if necessary. We might be able to combine common logic with raw_data processing: For comparison, the ElasticSearch PR is [here](https://github.com/grafana/grafana/pull/63647). 

Some general guidelines are:

* Create visualization to compare between frontend and backend flows, preferably in a shared Cloud dashboard e.g. clouddatasources.grafana.net 
* Unit tests in the frontend should be copied or updated in the backend to assert the same behavior and add unit tests when there are none.
    * Check Elasticsearch repo for related tests – helpful when there are no OpenSearch tests in the frontend or backend
* We should match OpenSearch's backend behavior to the current frontend behavior, but be aware it may not always make sense. Elasticsearch's current behavior is also another resource for how OpenSearch should behave.
    * Let's be pragmatic about this – our foremost goal is the migration of OpenSearch's frontend behavior. We can aim for some Elasticsearch feature parity, but only _within reason_.
* Often there is a PR in the Elasticsearch backend data source migration [https://github.com/grafana/grafana/issues/54011](https://github.com/grafana/grafana/issues/54011) which is 75% of what OpenSearch will need
* Migrate or implement: i.e. make any changes to code in backend
* Use the dashboard created above to compare visualization results (data frame comparison)
    * Can also inspect data frame on a more detailed basis
    
    
**Note:** It seems like we have a problem, at least on the FE, with detecting field types, for example here the Query inspector shows AvgTicketPrice as string and not int, which isn't what I would expect. Just something to keep in mind when comparing Logs. 

 
<img width="413" alt="Screenshot 2023-06-12 at 1 26 02 PM" src="https://github.com/grafana/opensearch-datasource/assets/16140639/85a9545e-48c8-4409-9f4d-2f75d1ba88e9">
<img width="286" alt="Screenshot 2023-06-12 at 1 28 30 PM" src="https://github.com/grafana/opensearch-datasource/assets/16140639/e47f76d6-da77-4d9c-a14c-357888ed177f">

PR #228, opened 2023-07-31
Migrate Lucene log queries to the backend in Explore view only

fridgepoet:
Lucene log queries in Explore have been refactored to execute through the backend (Go code) in order to align with other data sources and improve support for alerting and expressions.

This PR is based on the current code in the [Elasticsearch](https://github.com/grafana/grafana/tree/main/pkg/tsdb/elasticsearch).

Changes from frontend to backend flow:
* The "Level field name" `level` is no longer maintained the third field in the table view. This is the same change as in the Elasticsearch PR: `level` field on frontend was added to the third position (after timestamp and line), but this is not needed, as we detect it based on name of the field and not its position. This meant, that in log details, level field wasn't alphabetically sorted. As there is no dependency on position of level I decided to keep it's position based on alphabetical order.
* Returned data is now typed, e.g. string, boolean. The previous frontend flow typed fields other than timestamp as string only. 

**Lucene logs queries will only run in the backend in the Explore view** in order to limit this migration's impact to dashboards. Please [let us know](https://github.com/grafana/opensearch-datasource/issues/new?assignees=&labels=datasource%2FOpenSearch%2Ctype%2Fbug&projects=&template=bug_report.md) if you observe some unwanted behavior while executing Lucene Log queries in Explore.

**Testing notes**:
The current version of the plugin (frontend flow) can be tested on [this dashboard](https://clouddatasources.grafana.net/goto/dPURnXqVg?orgId=1). 

We've done a manual comparison of the returned data frames between a frontend flow and backend flow. 
For the same datasource and same time range, for frontend and for backend flow, we compared the copy of the returned data frames (query inspector -> json -> data frames). No changes in the returned values have been observed.

Contributes to https://github.com/grafana/opensearch-datasource/issues/199

**Special notes**
Original Elasticsearch PR https://github.com/grafana/grafana/pull/63647


pkg/opensearch/response_parser.go
```diff
@@ -119,8 +190,8 @@ func processRawDataResponse(res *es.SearchResponse, timeField string, queryRes b
 			}
 		}
 
-		if timestamp, ok := getTimestamp(hit, timeField); ok {
-			doc[timeField] = timestamp
+		if timestamp, ok := getTimestamp(hit, configuredFields.TimeField); ok {
```
fridgepoet, line 193: Diff view may look like it changes `processRawData`, but please note the added code is `processLogsResponse`. `processRawData` only uses the configured TimeField

pkg/opensearch/response_parser.go
```diff
@@ -98,7 +102,74 @@ func (rp *responseParser) getTimeSeries(configuredFields es.ConfiguredFields) (*
 	return result, nil
 }
 
-func processRawDataResponse(res *es.SearchResponse, timeField string, queryRes backend.DataResponse) backend.DataResponse {
+const level = "level"
```
sarahzinger, line 105: where do we use this?

pkg/opensearch/response_parser.go
```diff
@@ -27,8 +27,10 @@ const (
 	filtersType     = "filters"
```
sarahzinger, line 27: Random thought for a future pr: I'm wondering if we should break this file down a bit into smaller files? like would this be easier to read maybe if the different response parsers had their own files?

pkg/opensearch/response_parser_test.go
```diff
@@ -1988,10 +2344,46 @@ func TestProcessRawDocumentResponse(t *testing.T) {
...
-func Test_sortPropNames_puts_timeField_at_the_beginning(t *testing.T) {
-	actual := sortPropNames(
-		map[string]bool{"_id": true, "Average": true, "timestamp": true},
-		client.ConfiguredFields{TimeField: "timestamp"},
-	)
-	assert.Equal(t, []string{"timestamp", "Average", "_id"}, actual)
+func Test_sortPropNames(t *testing.T) {
+	t.Run("returns slice after finding fields in prop names and placing them in front", func(t *testing.T) {
+		actual := sortPropNames(
+			map[string]bool{"lookForThisField": true, "_another_field": true},
```
sarahzinger, line 2350: maybe slightly more obvious if we switch the order? `map[string]bool{"_another_field": true, "lookForThisField": true },`

sarahzinger (approved):
didn't do a manual test but ping me if you'd like me to! Looks good tho!

iwysiu (approved):
Looks good! some comments. I also did not run it, but can run it if you would like

idastambuk:
Pulled it and made some logs queries - looks good!

pkg/opensearch/response_parser.go
fridgepoet: It was used once down in line ~300-something. Not sure if you didn't like that since it was only used once? Removed it

pkg/opensearch/lucene_handler.go
```diff
@@ -61,13 +62,59 @@ func (h *luceneHandler) processQuery(q *Query) error {
...
 	return nil
 }
 
+func processLogsQuery(q *Query, b *es.SearchRequestBuilder, from, to int64, defaultTimeField string) {
+	metric := q.Metrics[0]
+	b.Sort(descending, defaultTimeField, "boolean")
+	b.Sort(descending, "_doc", "")
+	b.AddDocValueField(defaultTimeField)
+	b.AddTimeFieldWithStandardizedFormat(defaultTimeField)
+	sizeString := metric.Settings.Get("size").MustString()
```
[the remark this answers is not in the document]
fridgepoet: No, it is passed to us as a string which is not coerced by that method.
That was mistakenly MustInt before which created an issue https://github.com/grafana/opensearch-datasource/pull/210

katebrenner:
just pulled it and ran it and seamless experience!

PR #233, opened 2023-08-03
Update trackQuery to track non-dashboard queries

ivanahuckova:
In [Loki](https://github.com/grafana/grafana/blob/main/public/app/plugins/datasource/loki/tracking.ts#L161) and [Elasticsearch](https://github.com/grafana/grafana/blob/main/public/app/plugins/datasource/elasticsearch/tracking.ts#L119) we track only non-dashboard queries, as dashboards can have 100s of panels and can have refresh interval of `5s` which leads to huge number of tracked queries.

If we are tracking only Explore/Panel View/Others queries, we can get a better insight into what queries are USERS executing and not which ones are run in dashboards on repeat.

So I decided to create this suggestion PR for you to consider. 🙂

### DIFF
--- a/src/tracking.ts
+++ b/src/tracking.ts
@@ -3,6 +3,10 @@ import { reportInteraction } from '@grafana/runtime';
 import { LuceneQueryType, OpenSearchQuery, QueryType } from 'types';
 
 export function trackQuery(response: DataQueryResponse, queries: OpenSearchQuery[], app: string): void {
+  if (app === CoreApp.Dashboard || app === CoreApp.PanelViewer) {
+    return;
+  }
+
   for (const query of queries) {
     try {
       reportInteraction('grafana_opensearch_query_executed', {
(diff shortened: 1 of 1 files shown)


fridgepoet (approved):
@idastambuk what do you think?

sarahzinger (approved):
in cloudwatch we subscribe to a "dashboard loaded" event https://github.com/grafana/grafana/blob/main/public/app/plugins/datasource/cloudwatch/module.tsx#L21

which I *think* fires off once with all queries in a dashboard the first time it loads. Could be useful if we want to track how users are building dashboards? 

That said I think this change as is seems like a good improvement to me

svennergr:
I think back then the `DashboardLoadedEvent` used when subscribing to those events by e.g.

```
getAppEvents().subscribe<DashboardLoadedEvent<CloudWatchQuery>>(DashboardLoadedEvent, onDashboardLoadedHandler);
```

wasn't supported from all version that we supported in the OpenSearch datasource.

Issue #232, opened 2023-08-03
Queries with invalid "queryType" panicking if in backend query flow 


fridgepoet:
**What happened**:
First reported [here](https://github.com/grafana/opensearch-datasource/pull/231#issuecomment-1662200822)
Any query that may have been created unintentionally with an invalid query like `"queryType": "randomWalk"` in the dashboard JSON model which has been migrated to a backend query flow will fail.
The backend query flow started in v2.7.0 with raw data queries and now includes raw document queries.

These `"queryType": "randomWalk"` queries are running into a panic right [here in the backend](https://github.com/grafana/opensearch-datasource/blob/f1e19f3da32c76a6fa4aabb65fde5c9288481021/pkg/opensearch/time_series_query.go#L40)

Workarounds:

* open the dashboard JSON model and set `"queryType": "lucene"` or `"queryType": "PPL"` as needed
* revert to v2.6.2

**How to reproduce it (as minimally and precisely as possible)**:
1. Create a Raw Data or Raw Document visualization.
2. Modify the dashboard JSON manually to have `"queryType": "randomWalk"` in the panel.

**Environment**:
- Grafana version: v10
- Plugin version: v2.8.2

PR #234, opened 2023-08-04
Improve error handling for invalid queryType

fridgepoet:
**This will change the behavior from v2.6.2 and below. Any invalid `queryType`s will need to be updated via panel JSON or the query editor to avoid these errors.**

Both the backend and frontend code did not gracefully handle the situation where a panel's `queryType` was not `lucene` or `PPL`. 
This PR aims to improve that.
* In the backend, an error [will now be raised](https://github.com/grafana/opensearch-datasource/pull/234/files#diff-cd8a3334d960501322423406758ab863448d27e3719a73b9a48e435d6c532474R85-R88) when a `queryType` is invalid.
This error will be [propagated to the highest backend level](https://github.com/grafana/opensearch-datasource/pull/234/files#diff-198336eb5f01515289763c8a9525875720312f75bb93338969707466a45caa38R72-R79) where it will eventually surface to the user. 
* In the frontend, a [placeholder will substitute](https://github.com/grafana/opensearch-datasource/pull/234/files#diff-5a70dde9f9cfdc6be9d19a7d18786ffef6a10241887783131bdf226cee4f8674R14-R17) for an invalid query. Using that menu to choose a new query and saving will overwrite the invalid query. Together with the backend warning, that looks like: 
Screenshot 2023-08-24 at 18 10 00

Fixes https://github.com/grafana/opensearch-datasource/issues/232

This is a breaking change compared to v2.6.2 and below. This is because previously any invalid `queryType`s were changed to Lucene queries.


pkg/opensearch/opensearch.go
```diff
@@ -62,7 +63,24 @@ func (ds *OpenSearchDatasource) QueryData(ctx context.Context, req *backend.Quer
 	}
 
 	query := newTimeSeriesQuery(client, req, intervalCalculator)
-	response, err := query.execute()
+	response, err := checkError(query.execute())
+	return response, err
+}
+
+func checkError(response *backend.QueryDataResponse, err error) (*backend.QueryDataResponse, error) {
```
iwysiu, line 70: nit: would `wrapError` make more sense as a name?

pkg/opensearch/opensearch.go
```diff
@@ -62,7 +63,24 @@ func (ds *OpenSearchDatasource) QueryData(ctx context.Context, req *backend.Quer
...
+	var invalidQueryTypeError invalidQueryTypeError
+	if errors.As(err, &invalidQueryTypeError) {
+		return &backend.QueryDataResponse{
+			Responses: map[string]backend.DataResponse{
+				invalidQueryTypeError.refId: {
+					Error: fmt.Errorf(`invalid queryType, should be Lucene or PPL`),
```
iwysiu, line 76: I'm not sure its important, but I think we're losing what the queryType actually was here, it looks like the error from `time_series_query.go` returns it as part of the error string

iwysiu (approved):

Issue #200, opened 2023-06-12
Backend: Migrate PPL Logs queries to the backend


idastambuk:
Prerequisite: https://github.com/grafana/opensearch-datasource/issues/226

PPL Logs are processed into dataFrames on the FE, and as timeSeries on the BE. They should probably be separated out.
ES plugin doesn't support PPL queries, but maybe the [Logs PR](https://github.com/grafana/grafana/pull/63647) for Lucene could be something to look at for reference.

Some general guidelines are:

* Create visualization to compare between frontend and backend flows, preferably in a shared Cloud dashboard e.g. clouddatasources.grafana.net 
* Unit tests in the frontend should be copied or updated in the backend to assert the same behavior and add unit tests when there are none.
    * Check Elasticsearch repo for related tests – helpful when there are no OpenSearch tests in the frontend or backend
* We should match OpenSearch's backend behavior to the current frontend behavior, but be aware it may not always make sense. Elasticsearch's current behavior is also another resource for how OpenSearch should behave.
    * Let's be pragmatic about this – our foremost goal is the migration of OpenSearch's frontend behavior. We can aim for some Elasticsearch feature parity, but only _within reason_.
* Often there is a PR in the Elasticsearch backend data source migration [https://github.com/grafana/grafana/issues/54011](https://github.com/grafana/grafana/issues/54011) which is 75% of what OpenSearch will need
* Migrate or implement: i.e. make any changes to code in backend
* Use the dashboard created above to compare visualization results (data frame comparison)
    * Can also inspect data frame on a more detailed basis
   
  
Note: ([Similarly](https://github.com/grafana/opensearch-datasource/issues/199) to Lucene Logs of the FE) It looks like PPL Logs are not completely correctly processed on the Frontend, as the int field is passed as string. This isn't the case on the BE, as seen from the screenshot. 
<img width="500" alt="Screenshot 2023-06-12 at 1 31 26 PM" src="https://github.com/grafana/opensearch-datasource/assets/16140639/2a3ec0da-f082-4b87-8de4-048367cf78a5">
<img width="300" alt="Screenshot 2023-06-12 at 1 32 18 PM" src="https://github.com/grafana/opensearch-datasource/assets/16140639/2c801a51-312b-4311-8551-c4314ac7457a">

sarahzinger:
maybe we can also fix this one: https://github.com/grafana/opensearch-datasource/issues/212 with this ticket

PR #244, opened 2023-08-28
Apply ad-hoc filters to PPL queries before sending it to the backend

### LINKED ISSUE #226: Frontend: Pre-requisite for PPL migrations - refactor adhoc filters on frontend before sending query to backend
Similar to https://github.com/grafana/opensearch-datasource/pull/225, but for PPL

kevinwcyu:

Prerequisite work for migrating PPL queries to be executed on the backend. Ad-hoc filters are applied on the frontend only so this work ensures they're added before sending a query to the backend.

Fixes #226 

You currently can't test this except for running the unit tests since no PPL queries have been migrated to execute on the backend yet.

### DIFF
--- a/src/datasource.test.ts
+++ b/src/datasource.test.ts
@@ -1584,37 +1584,75 @@ describe('OpenSearchDatasource', function(this: any) {
     });
   });
 
-  describe('addAdhocFilters', () => {
+  describe('addAdHocFilters', () => {
     describe('with invalid filters', () => {
-      it('should filter out ad hoc filter without key', () => {
-        const query = ctx.ds.addAdHocFilters('foo:"bar"', [{ key: '', operator: '=', value: 'a', condition: '' }]);
-        expect(query).toBe('foo:"bar"');
-      });
-      it('should filter out ad hoc filter without value', () => {
-        const query = ctx.ds.addAdHocFilters('foo:"bar"', [{ key: 'a', operator: '=', value: '', condition: '' }]);
-        expect(query).toBe('foo:"bar"');
+      describe('Lucene queries', () => {
+        it('should filter out ad hoc filter without key', () => {
+          const query = ctx.ds.addAdHocFilters({ refId: 'A', query: 'foo:"bar"', queryType: QueryType.Lucene }, [
+            { key: '', operator: '=', value: 'a', condition: '' },
+          ]);
+          expect(query).toBe('foo:"bar"');
+        });
+
+        it('should filter out ad hoc filter without value', () => {
+          const query = ctx.ds.addAdHocFilters({ refId: 'A', query: 'foo:"bar"', queryType: QueryType.Lucene }, [
+            { key: 'a', operator: '=', value: '', condition: '' },
+          ]);
+          expect(query).toBe('foo:"bar"');
+        });
+
+        it('should filter out filter ad hoc filter with invalid operator', () => {
+          const query = ctx.ds.addAdHocFilters({ refId: 'A', query: 'foo:"bar"', queryType: QueryType.Lucene }, [
+            { key: 'a', operator: 'A', value: '', condition: '' },
+          ]);
+          expect(query).toBe('foo:"bar"');
+        });
       });
 
-      it('should filter out filter ad hoc filter with invalid operator', () => {
-        const query = ctx.ds.addAdHocFilters('foo:"bar"', [{ key: 'a', operator: 'A', value: '', condition: '' }]);
-        expect(query).toBe('foo:"bar"');
+      describe('PPL queries', () => {
+        it('should filter out ad hoc filter without key', () => {
+          const query = ctx.ds.addAdHocFilters({ refId: 'A', query: 'source = test-index', queryType: QueryType.PPL }, [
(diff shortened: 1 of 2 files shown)


src/datasource.test.ts
```diff
@@ -1584,60 +1584,123 @@ describe('OpenSearchDatasource', function(this: any) {
...
-        expect(query).toBe('bar:"baz" AND -job:"grafana" AND service:/service/ AND count:>1');
+      describe('PPL queries', () => {
+        const adHocFilters = [
+          { key: 'bar', operator: '=', value: 'baz', condition: '' },
+          { key: 'job', operator: '!=', value: 'grafana', condition: '' },
+          { key: 'bytes', operator: '>', value: 50, condition: '' },
+          { key: 'count', operator: '<', value: 100, condition: '' },
+          { key: 'timestamp', operator: '=', value: '2020-11-22 16:40:43', condition: '' },
+        ];
+        it('should return query with ad-hoc filters applied', () => {
```
iwysiu, line 1695: Should this also have the query is empty case or is that not possible for ppl queries?

iwysiu (approved):
Approved with a question about ppl queries

src/datasource.test.ts
```diff
@@ -1584,60 +1584,123 @@ describe('OpenSearchDatasource', function(this: any) {
...
+          const query = ctx.ds.addAdHocFilters({ refId: 'A', query: 'source = test-index', queryType: QueryType.PPL }, [
+            { key: 'a', operator: '=~', value: '', condition: '' },
+            { key: 'a', operator: '!~', value: '', condition: '' },
+          ]);
+          expect(query).toBe('source = test-index');
+        });
       });
     });
 
     describe('with 1 ad hoc filter', () => {
```
katebrenner, line 1637: just curious - any reason why only testing Lucene in this test case?

katebrenner (approved):

src/datasource.test.ts
kevinwcyu: i didn't think there was anything that could be added under this test case with 1 ad hoc filter for PPL queries that couldn't be covered by the multiple ad hoc filters test cases.

src/datasource.test.ts
kevinwcyu: right now the flow of the PPL execution is to always make sure the query string is not empty by adding the index that is being searched to the query string if the query string is empty

https://github.com/grafana/opensearch-datasource/blob/6e6f9affd5bfbf66ac939a90768782dc7b176ce3/src/datasource.ts#L741-L743

this happens happens before ad-hoc filters are added to the query string, so for this PR, i'm just testing the adding of the ad-hoc filters.

an empty query string test case will probably be more suited for https://github.com/grafana/opensearch-datasource/issues/200

in short, ppl queries should not be empty, but that should be handled outside of the `addAdHocFilters` code.

src/datasource.ts
```diff
@@ -396,6 +404,31 @@ export class OpenSearchDatasource extends DataSourceWithBackend<OpenSearchQuery,
     return [query, ...osFilters].filter(f => f).join(' AND ');
   }
 
+  addPPLAdHocFilters(queryString: string, adHocFilters: any) {
+    for (let i = 0; i < adHocFilters.length; i++) {
+      const { key, operator } = adHocFilters[i];
+      let { value } = adHocFilters[i];
+
+      if ('=~' === operator || '!~' === operator || !key || !value) {
```
fridgepoet, line 412: Would you explain more about this condition? What's that about the operators? Was this already illustrated in a test case and I missed it?

src/datasource.ts
```diff
@@ -396,6 +404,31 @@ export class OpenSearchDatasource extends DataSourceWithBackend<OpenSearchQuery,
     return [query, ...osFilters].filter(f => f).join(' AND ');
   }
 
+  addPPLAdHocFilters(queryString: string, adHocFilters: any) {
+    for (let i = 0; i < adHocFilters.length; i++) {
+      const { key, operator } = adHocFilters[i];
```
fridgepoet, line 409: Just curious: I notice in the existing `addPPLAdhocFilters` in `buildPPLQuery` that they're always using `adhocFilters[i].value`. How come we don't have to use `.value` here?

fridgepoet (approved):
Nice!

src/datasource.ts
kevinwcyu: this wasn't in a test case before. these two operators are not recognized by opensearch so it returns an error if you try to use them.

src/datasource.ts
kevinwcyu: using `let { value } = adHocFilters[i];` is shorter and easier to read in the later code since you can just use `value` instead of `adhocFilters[i].value`.

src/datasource.ts
fridgepoet: are there tests now documenting this? if not, would it be worth it to add them? honestly asking that and deferring to your judgment about this!

kevinwcyu: https://github.com/grafana/opensearch-datasource/pull/244/files#diff-f5855f08029d58a7871f69a0ebfe07376b2508725c81159df890fc793eb5b545R1629-R1630 there's this test case

Issue #250, opened 2023-09-05
OpenSearch: Incorrectly creates new HTTP client on each query


fridgepoet:
**What happened**:
Similar to what was reported in Elasticsearch https://github.com/grafana/grafana/issues/54233

A new HTTP client is created on each query, [here](https://github.com/grafana/opensearch-datasource/blob/b2812efd4b20bafbccf0fa3378b96fb57f2f0b43/pkg/opensearch/client/client.go#L323) for example. This is not good since HTTP connections can't be reused and could in theory use up all available TCP sockets on a server/node.

**What you expected to happen**:
That it creates a new HTTP client for a non-instantiated datasource instance and reuses that throughout the life-cycle of that instance.

PR #252, opened 2023-09-06
Reuse http Client in the backend

fridgepoet:
Similar to the fix in Elasticsearch: https://github.com/grafana/grafana/pull/55172

The http.Client was initialized with every [Lucene](https://github.com/grafana/opensearch-datasource/blob/b2812efd4b20bafbccf0fa3378b96fb57f2f0b43/pkg/opensearch/client/client.go#L323) and [PPL query execution](https://github.com/grafana/opensearch-datasource/blob/b2812efd4b20bafbccf0fa3378b96fb57f2f0b43/pkg/opensearch/client/client.go#L546).

This PR is a refactor so that the http.Client is initialized at a higher level. This Client is passed to the entry point of query execution so that it can be reused. 

Fixes #https://github.com/grafana/opensearch-datasource/issues/250

The following types of queries were tested for non-regression: Lucene raw_data and raw_document queries are executed in the backend. Lucene logs queries are executed on the backend in Explore.
An AWS Serverless instance was also tested with SigV4 authentication.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,11 @@
 
 All notable changes to this project will be documented in this file.
 
-## 2.10.0
+## 2.10.1
 
+- Backend: Refactor http client so that it is reused
+
+## 2.10.0
 
 - [Explore] Migrate Lucene log queries to the backend by @fridgepoet in https://github.com/grafana/opensearch-datasource/pull/228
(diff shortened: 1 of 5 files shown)


sarahzinger (approved):
Lgtm!

Issue #223, opened 2023-07-24
Backend: Implement Lucene Traces in the backend


sarahzinger:
There's currently no processing of traces on the backend, only frontend. There is no equivalent PR in ElasticSearch to reference, but we have implemented this fairly recently so there are up-to-date tests and the feature looks stable. 

```[tasklist]
### Tasks
- [x] #280
- [ ] https://github.com/grafana/opensearch-datasource/issues/287
- [ ] https://github.com/grafana/opensearch-datasource/issues/303
```

Note: This is the case for jaeger trace format, as otel hasn't been implemented yet (https://github.com/grafana/opensearch-datasource/issues/205). However, the scope of this PR is just migrating the data processing for jaeger traces that we currently have in FE and in data-prepper.

Some general guidelines are:

* Create visualization to compare between frontend and backend flows, preferably in a shared Cloud dashboard e.g. clouddatasources.grafana.net 
* Unit tests in the frontend should be copied or updated in the backend to assert the same behavior and add unit tests when there are none.
    * Check Elasticsearch repo for related tests – helpful when there are no OpenSearch tests in the frontend or backend
* We should match OpenSearch's backend behavior to the current frontend behavior, but be aware it may not always make sense. Elasticsearch's current behavior is also another resource for how OpenSearch should behave.
    * Let's be pragmatic about this – our foremost goal is the migration of OpenSearch's frontend behavior. We can aim for some Elasticsearch feature parity, but only _within reason_.
* Migrate or implement: i.e. make any changes to code in backend
* Use the dashboard created above to compare visualization results (data frame comparison)
    * Can also inspect data frame on a more detailed basis

PR #257, opened 2023-09-13
Backend: Refactor trace spans (query building + response processing)

idastambuk:
Implements https://github.com/grafana/opensearch-datasource/issues/223

Support for trace spans on the backend, only in Explore, to follow our rollout stage plan. 

Screenshot 2023-10-19 at 8 38 31 PM

Fixes https://github.com/grafana/opensearch-datasource/issues/223

I couldn't get the snapshot test to work, because in response processing I worked with a lot of maps, which makes the output not follow consistent order of fields. This doesn't matter when it comes to the trace visualization, as it needs specific named fields to display different parts of the UI. Therefore I decided I wouldn't change the original code in order to comply with tests (like sorting fields etc), but added a few utils int he tests to just check that all the fields are in the output, wherever.


src/traces/formatTraces.ts
```diff
@@ -188,8 +188,7 @@ function getStackTraces(events: OpenSearchSpanEvent[]): string[] | undefined {
   const stackTraces = events
     .filter(event => event.attributes.error)
     .map(event => `${event.name}: ${event.attributes.error}`);
-  // if we return an empty array, Trace panel plugin shows "0"
-  return stackTraces.length > 0 ? stackTraces : undefined;
+  return stackTraces;
```
idastambuk, line 191: fixed this in the Trace plugin a while ago, so Im updating this.

src/datasource.ts
```diff
@@ -543,7 +543,8 @@ export class OpenSearchDatasource extends DataSourceWithBackend<OpenSearchQuery,
               metric.type === 'raw_document' ||
               (request.app === CoreApp.Explore && target.queryType === QueryType.Lucene)
           ) ||
-          (request.app === CoreApp.Explore && target.queryType === QueryType.PPL && target.format === 'logs')
+          (request.app === CoreApp.Explore && target.queryType === QueryType.PPL && target.format === 'logs') ||
+          (request.app === CoreApp.Explore && target.queryType === QueryType.Lucene)
```
fridgepoet, line 547: Can we limit this to just the trace spans as the trace list is not yet fully in the backend? thanks!

pkg/opensearch/lucene_handler.go
```diff
@@ -79,6 +91,18 @@ func (h *luceneHandler) processQuery(q *Query) error {
 	return nil
 }
 
+func getTraceId(rawQuery string) string {
+	trimmed := strings.TrimSpace(rawQuery)
+	re := regexp.MustCompile(`traceId:(.+)`)
+	matches := re.FindStringSubmatch(trimmed)
```
iwysiu, line 97: What happens if something like `foo traceId:test` is passed? Does it get parsed successfully, and do we want that to happen?

pkg/opensearch/snapshot_tests/lucene_trace_spans_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"net/http"
+	"testing"
+
+	"github.com/grafana/grafana-plugin-sdk-go/backend"
+	"github.com/grafana/opensearch-datasource/pkg/opensearch"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_trace_spans_request(t *testing.T) {
```
iwysiu, line 15: should there be a response test?

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.20
 require (
 	github.com/Masterminds/semver v1.5.0
 	github.com/bitly/go-simplejson v0.5.0
+	github.com/envoyproxy/go-control-plane v0.10.3
```
iwysiu, line 8: what is this being added for? I couldn't find where it was being used, but I might have missed it.

pkg/opensearch/response_parser.go
```diff
@@ -86,29 +87,187 @@ func (rp *responseParser) getTimeSeries(configuredFields es.ConfiguredFields) (*
...
+			"_index":  hit["_index"],
+			"_source": withKeysToObj,
+		}
+
+		for k, v := range withKeysToObj {
+			// some field names TraceView viz needs do not correspond to what we get from OpenSearch, this remaps them
+			if k == "startTime" {
+				startTime, err := utils.TimeFieldToMilliseconds(v)
+				if err != nil {
+					return backend.ErrDataResponse(500, err.Error())
```
iwysiu, line 143: Should we wrap this error so we know its from `startTime`?

pkg/opensearch/response_parser.go
```diff
@@ -86,29 +87,187 @@ func (rp *responseParser) getTimeSeries(configuredFields es.ConfiguredFields) (*
...
+		}
+
+		doc := map[string]interface{}{
+			"_id":     hit["_id"],
+			"_type":   hit["_type"],
+			"_index":  hit["_index"],
+			"_source": withKeysToObj,
+		}
+
+		for k, v := range withKeysToObj {
```
iwysiu, line 138: is this missing `serviceName`?

As a style nit, I think you could do a `switch` on the value of `k`

pkg/opensearch/response_parser.go
```diff
@@ -86,29 +87,187 @@ func (rp *responseParser) getTimeSeries(configuredFields es.ConfiguredFields) (*
...
+				doc["spanID"] = v
+				continue
+			}
+			if k == "name" {
+				doc["operationName"] = v
+				continue
+			}
+			if k == "resource" {
+				resourceAttributes, ok := v.(map[string]interface{})["attributes"].(map[string]interface{})
+				if resourceAttributes != nil && ok {
```
iwysiu, line 177: It seems like we could separate out a utility function for converting the key values. A helper for the `events` would also help reduce the nesting and make it easier to read.

pkg/opensearch/response_parser.go
```diff
@@ -86,29 +87,187 @@ func (rp *responseParser) getTimeSeries(configuredFields es.ConfiguredFields) (*
...
+	frame := data.NewFrame("", fields...)
+	if frame.Meta == nil {
+		frame.Meta = &data.FrameMeta{}
+	}
+	frame.Meta.PreferredVisualization = data.VisTypeTrace
+
+	queryRes.Frames = data.Frames{frame}
+	return queryRes
+}
+func processLogsResponse(res *es.SearchResponse, limitString string, configuredFields es.ConfiguredFields, queryRes backend.DataResponse) backend.DataResponse {
```
iwysiu, line 270: I don't think `limitString` is being used anywhere.

pkg/utils/utils.go
```diff
@@ -30,4 +33,61 @@ func NullFloatToNullableTime(ts null.Float) *time.Time {
...
+		}
+
+		current[keys[len(keys)-1]] = value
+	}
+
+	return result
+}
+
+func TimeFieldToMilliseconds(date interface{}) (int64, error) {
+	layout := "2006-01-02T15:04:05.000000Z"
```
iwysiu, line 59: style nit: You could move this to a const at the top of the file

pkg/utils/utils.go
```diff
@@ -30,4 +33,61 @@ func NullFloatToNullableTime(ts null.Float) *time.Time {
...
+	case int64:
+		timestamp = &timeField
+	default:
+		return 0, errors.New("unrecognized time format")
+	}
+	return *timestamp, nil
+}
+func SpanHasError(spanEvents []interface{}) bool {
+	for _, event := range spanEvents {
+		if eventMap, ok := event.(map[string]interface{}); ok {
```
iwysiu, line 78: also obnoxious style nit: You could do
```
eventMap, ok := event.(map[string]interface{})
if !ok {
   log.DefaultLogger.Debug("span event is not a map")
   continue
}
attributes := eventMap["attributes"]
...
```
(The idea behind the style nit is returning/continuing early and reducing indents)

pkg/opensearch/snapshot_tests/lucene_trace_spans_test.go
idastambuk: Sadly I couldn't get the response test to work in snapshot because the response processing uses maps, so the result has slightly different order every time. Snapshot tests fail for this reason. Will leave a comment though, good catch!

pkg/opensearch/lucene_handler.go
idastambuk: Before and after the refactor we've had it pass successfully and take into account just the trace id - we don't really support any additional filtering in the query field. Im ok with leaving it live that and not changing the functionality

pkg/opensearch/response_parser.go
```diff
@@ -86,29 +87,187 @@ func (rp *responseParser) getTimeSeries(configuredFields es.ConfiguredFields) (*
...
-		default:
-			props := make(map[string]string)
-			err := rp.processBuckets(res.Aggregations, target, &queryRes, props, 0)
-			if err != nil {
-				return nil, err
+		// trace span condition
+		if target.luceneQueryType == "Traces" {
+			queryRes = processTraceSpansResponse(res, queryRes)
+		} else {
+			switch target.Metrics[0].Type {
```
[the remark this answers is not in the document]
idastambuk: The reason there's an `else` here is that for some reason trace queries are sent from the FE with a metrics field, so the switch would overwrite the response from the traces query. 
This should definitely be fixed on the FE in a future refactor. 
I will refactor so it returns the traces response early, which avoids the nested switch, hope that looks better!

go.mod
idastambuk: Don't remember, but it's definitely not used 👀 removed it!

pkg/opensearch/response_parser.go
idastambuk: The `serviceName` field stays the same, as it comes from OpenS in the correct string we need and we don't need to remap it.

pkg/opensearch/response_parser.go
idastambuk: Fixed it after rebase!

iwysiu (commented):
The changes look good! I added a few more small comments.

As a heads up, if you rebase and force push a PR that already has comments on it, it messes up the history and breaks the links for all the old comments, which makes it harder to review. You can do a merge commit instead to fix the merge conflicts.

pkg/opensearch/response_parser_test.go
```diff
@@ -2341,3 +2342,242 @@ func Test_sortPropNames(t *testing.T) {
...
+	assert.Equal(t, sortedLogs[0].Timestamp, 1697615918486)
+	require.Equal(t,2, len(sortedLogs))
+	
+}
+
+type KeyValue struct {
+	Key   string
+	Value any
+}
+func sortObjectsByKey(rawObject *data.Field) []KeyValue {
```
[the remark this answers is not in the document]
idastambuk: 😲 Good to know!

pkg/opensearch/response_parser.go
```diff
@@ -109,6 +118,144 @@ func (rp *responseParser) getTimeSeries(configuredFields es.ConfiguredFields) (*
...
+				{
+					doc["operationName"] = v
+					continue
+				}
+			case "resource":
+				{
+					resourceAttributes, ok := v.(map[string]interface{})["attributes"].(map[string]interface{})
+					if resourceAttributes != nil && ok {
+						transformedResourceAttributes := getKeyTraceSpanValuePairs(resourceAttributes)
+						if transformedResourceAttributes != nil {
```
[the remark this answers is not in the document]
idastambuk: It isn't, good catch!

pkg/opensearch/response_parser.go
```diff
@@ -87,6 +88,14 @@ func (rp *responseParser) getTimeSeries(configuredFields es.ConfiguredFields) (*
 			Frames: data.Frames{},
 		}
 
+		// trace span condition
+		// trace queries are sent from the FE with a metrics field, returning early so the switch doesn't overwrite the response from the traces query
+		if target.luceneQueryType == luceneQueryTypeTraces {
+			queryRes = processTraceSpansResponse(res, queryRes)
+			result.Responses[target.RefID] = queryRes
+			return result, nil
```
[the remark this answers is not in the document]
idastambuk: You're absolutely right, it should be a continue

iwysiu (approved):
Nice work on all the go code! Approved with one incredibly nitpicky style comment.

PR #265, opened 2023-09-20
Get filter values with correct time range

### LINKED ISSUE #213: Timestamp filter values aren't being fetched correctly
**What happened**:
When fetching filter values, we call [getTerms](https://github.com/grafana/opensearch-datasource/blob/6718c3779868d451a9b0461455fbaa9f632cc114/src/datasource.ts#L873C8-L873C8) without a range value, which means that when timestamp values are fetched they're always from the default range.

I expect it to fetch all and only the values from the range

Create a PPL query (local traces with the ppl query `source = security-auditlog-2023.07.12` but replace the auditlog date with a day with data) and click filter for value for a timestamp value. If you open the filter dropdown/check the network tab, you should see that it's using the default range.

Even after fixing this, filtering on a timestamp still won't work until https://github.com/grafana/opensearch-datasource/issues/212 is fixed.

**Environment**:
- Grafana version:
- Plugin version:

iwysiu:
Depends on https://github.com/grafana/grafana/pull/74952 or something similar that actually passes the timeRange to getTagValues to work. (If timeRange isn't passed then nothing changes)
This passes the timeRange of the getTagValue query along to getTerms, so we'll get values in the correct range instead of the default last 6 hours

Fixes #213

This has to be run with a version of main grafana with the above pr merged for the timeRange field to actually be passed in

### DIFF
--- a/src/datasource.ts
+++ b/src/datasource.ts
@@ -964,7 +964,7 @@ export class OpenSearchDatasource extends DataSourceWithBackend<OpenSearchQuery,
   }
 
   getTagValues(options: any) {
-    return this.getTerms({ field: options.key, query: '*' });
+    return this.getTerms({ field: options.key, query: '*' }, options.timeRange);
   }
 
   targetContainsTemplate(target: any) {


kevinwcyu (approved):
tested with main grafana and works nicely

idastambuk (approved):

Issue #197, opened 2023-06-08
Backend: Migrate Lucene metrics to the backend


idastambuk:
Note: Blocked by https://github.com/grafana/opensearch-datasource/issues/224

Currently, Lucene metric queries are processed as [time series](https://github.com/grafana/opensearch-datasource/blob/ace67550f9591b0298488aac2044be96b30c89e4/src/OpenSearchResponse.ts#L524) (frontend). This flow needs to be compared and migrated to the backend. Some of the logic already exists on the backend [here](https://github.com/grafana/opensearch-datasource/blob/ace67550f9591b0298488aac2044be96b30c89e4/pkg/opensearch/response_parser.go#L45), but it might need to be updated. 

-------
To tackle this issue, we'll start by comparing the way the backend and frontend:
- build the query to OpenSearch
- return the response in Grafana

Differences between frontend/backend:
- Metric, Metric Max, Group By Terms
  - Query building to OpenSearch
    - [x] In `terms`,  `order`'s key is `_term` in backend, `_key` in frontend https://github.com/grafana/opensearch-datasource/pull/270
    - [x] `min_doc_count` missing in backend, present in frontend https://github.com/grafana/opensearch-datasource/pull/268
    - [x] `size` is 500 in backend, 0 in frontend https://github.com/grafana/opensearch-datasource/pull/269
  - Grafana data frames response:
    - Backend has: additional `typeInfo` object for `AvgTicketPrice` and `Max` with `"typeInfo": {
              "frame": "float64",
              "nullable": true`
    - Frontend does not have `typeInfo`
    - Conclusion: Backend seems acceptable as this is how the equivalent Elasticsearch query also returns data frames.
- Metric Sum / Group By Date Histogram / Alias
  - Query building to OpenSearch
    - no difference between backend and frontend
  - Grafana data frames response:
    - Backend has additional "config" object with "displayNameFromDS": "distanceKms" : this is expected due to https://github.com/grafana/opensearch-datasource/pull/180. Frontend has an additional `"name": "distanceKms"` object which is replaced by the `displayNameFromDS` in the backend.  This is also the same behavior in Elasticsearch for the same type of query.
    - Similar to above, backend has additional `typeInfo` object
    - Backend has additional `"meta": {
          "type": "timeseries-multi",
          "typeVersion": [
            0,
            0
          ]
[...]

PR #272, opened 2023-10-03
Explore: Execute Lucene metric queries through the backend 

fridgepoet:
In Grafana Explore, metric aggregations such as Count, Sum, Min, Max, Average, Extended Stats, Percentiles, and Unique Count (and pipeline aggregations) are now executed through the backend.

**Queries with these metric aggregations will only run in the backend in the Explore view** in order to limit this migration's impact to dashboards. Please [let us know](https://github.com/grafana/opensearch-datasource/issues/new?assignees=&labels=datasource%2FOpenSearch%2Ctype%2Fbug&projects=&template=bug_report.md) if you observe any unwanted behavior.

Closes https://github.com/grafana/opensearch-datasource/issues/197

Differences between backend and frontend data frames:
- Backend data frames have additional `"typeInfo": {
              "frame": "float64",
              "nullable": true` This seems acceptable as this is how the equivalent Elasticsearch query also returns data frames.
- For a query with Alias, backend data frames have additional "config" object with "displayNameFromDS": "distanceKms". This is expected due to https://github.com/grafana/opensearch-datasource/pull/180. This is also the same behavior in Elasticsearch for the same type of query.
- Backend data frames have additional `"meta": {
          "type": "timeseries-multi",
          "typeVersion": [
            0,
            0
          ]
        }` object. This is due to code based on an Elasticsearch PR https://github.com/grafana/grafana/pull/60169 and was included in https://github.com/grafana/opensearch-datasource/pull/180 and https://github.com/grafana/opensearch-datasource/pull/168. This is also the same behavior in Elasticsearch for the same type of query.
- Backend has additional ` "labels": {
              "metricId": "1"
            }` object. The presence and use of this object is consistent with Elasticsearch for the same type of query.

[Full differences and actions taken available in the description of the issue](https://github.com/grafana/opensearch-datasource/issues/197)


idastambuk:
Impressive investigation @fridgepoet! I clicked around running different types of queries and all looks great!

idastambuk (approved):

src/datasource.ts
```diff
@@ -541,7 +541,7 @@ export class OpenSearchDatasource extends DataSourceWithBackend<OpenSearchQuery,
             metric =>
               metric.type === 'raw_data' ||
               metric.type === 'raw_document' ||
-              (request.app === CoreApp.Explore && target.queryType === QueryType.Lucene && metric.type === 'logs')
+              (request.app === CoreApp.Explore && target.queryType === QueryType.Lucene && metric.type)
```
kevinwcyu, line 544:
```suggestion
              (request.app === CoreApp.Explore && target.queryType === QueryType.Lucene)
```

now that we are sending all lucene explore queries to the backend, do we need the check for metric.type here at all? would it ever be undefined, and if it is, would it matter? this is just a minor nit.

kevinwcyu (approved):
nice outline of the differences and explanations 🎉

src/datasource.ts
fridgepoet: Was wondering this myself, tried it out locally. Didn't see anything weird.
I'll merge with your suggestion and continue keeping an eye out. Thank yoou!

PR #277, opened 2023-10-05
Backend: Fix Lucene logs so it only uses date_histogram

fridgepoet:
We discovered that [in the frontend](https://github.com/grafana/opensearch-datasource/blob/c4bda8c85a56dc911b42aa8ad3f63079557a9058/src/datasource.ts#L749), Lucene logs queries were built only with `date_histogram` grouping by default time field, no matter what is chosen in the UI.
This fixes the backend to do the same. (This should only affect queries executed in Explore as of [v2.10.0](https://github.com/grafana/opensearch-datasource/releases/tag/v2.10.0).

Contributes to https://github.com/grafana/opensearch-datasource/issues/199

### DIFF
--- a/pkg/opensearch/snapshot_tests/testdata/lucene_logs.query_input.json
+++ b/pkg/opensearch/snapshot_tests/testdata/lucene_logs.query_input.json
@@ -7,14 +7,13 @@
     "alias": "",
     "bucketAggs": [
       {
-        "field": "timestamp",
+        "field": "agent.keyword",
         "id": "2",
         "settings": {
-          "interval": "20m",
-          "min_doc_count": "0",
-          "trimEdges": "0"
+          "interval": "1000",
+          "min_doc_count": "15"
         },
-        "type": "date_histogram"
+        "type": "histogram"
       }
     ],
     "format": "table",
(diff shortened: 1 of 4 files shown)


pkg/opensearch/snapshot_tests/testdata/lucene_logs.query_input.json
```diff
@@ -7,14 +7,13 @@
...
         "id": "2",
         "settings": {
-          "interval": "20m",
-          "min_doc_count": "0",
-          "trimEdges": "0"
+          "interval": "1000",
+          "min_doc_count": "15"
         },
-        "type": "date_histogram"
+        "type": "histogram"
```
fridgepoet, line 16: Test now shows that even with "histogram" and "min_doc_count" set, query is built with default date_histogram

pkg/opensearch/snapshot_tests/testdata/lucene_logs.query_input.json
```diff
@@ -7,14 +7,13 @@
     "alias": "",
     "bucketAggs": [
       {
-        "field": "timestamp",
```
iwysiu, line 10: is there a reason that the field changed?

iwysiu (approved):

pkg/opensearch/snapshot_tests/testdata/lucene_logs.query_input.json
fridgepoet: yes, sorry this should have also gone into an explanatory comment. this also shows that no matter the field put here, the "defaultTimeField" will be used over here https://github.com/grafana/opensearch-datasource/pull/277/files/98a8fa01a4bf2aa0adc11573355c0143f38a1c99#diff-1a311aa56365fe4dc3d664446f14d8c7835626e6a2a908be02eee3d403ad4808R44.

PR #286, opened 2023-10-17
Migrate to create-plugin and support node 18

kevinwcyu:
- Tested with Node v18.17.1 locally with the following `rm -rf node_modules && yarn && yarn lint && yarn test:ci && yarn build` then tested it with a simple query

Notes to reviewers:

I suggest checking the `Hide whitespace` checkbox in the diff view settings for this.

**Why does this include migrating from `@grafana/toolkit` to `@grafana/create-plugin`?** When updating to Node 18, we run into an error due to an incompatible version of node when running `yarn install` due to dependencies used in `@grafana/toolkit`, which is now deprecated. 

**What are the notable changes**
- Files created/modified by running `npx @grafana/create-plugin@latest migrate`
  - `./config/*` all generated files with default configuration for tooling such as webpack, eslint, jest, etc. The following root level files were also modified automatically to use the configuration defined in the `/.config` dir
    - `.eslintrc`
    - `.prettierrc.js`
    - `jest-setup.js`
    - `jest.config.js`
    - `tsconfig.json `
  - `package.json` modified to include required dev deps and updates the scripts
  - `docker-compose.yaml` used by `yarn server`
- `.github/workflows/detect-breaking-changes.yml` sets it to use the .nvmrc file to determine the version of node to use instead of hardcoding it
- `.nvmrc` is used to determine which version of Node is used to build the app in Drone.
- `CONTRIBUTING.MD` removes the `yarn watch` command which is no longer available, it is now just `yarn dev`
- `src/**/*.{ts,tsx}` files: formatting changes (prettier), and typescript fixes


.config/jest/mocks/react-inlinesvg.tsx
```diff
@@ -0,0 +1,25 @@
+// Due to the grafana/ui Icon component making fetch requests to
```
sarahzinger, line 1: lol this is so random

sarahzinger (approved):
holy moly what a journey! nice work!

PR #289, opened 2023-10-25
PPL: Execute Explore PPL Table format queries through the backend

### LINKED ISSUE #201: Backend: Migrate PPL Table Format to the backend
Prerequisite: https://github.com/grafana/opensearch-datasource/issues/226

Some general guidelines are:

* Create visualization to compare between frontend and backend flows, preferably in a shared Cloud dashboard e.g. clouddatasources.grafana.net 
* Unit tests in the frontend should be copied or updated in the backend to assert the same behavior and add unit tests when there are none.
    * Check Elasticsearch repo for related tests – helpful when there are no OpenSearch tests in the frontend or backend
* We should match OpenSearch's backend behavior to the current frontend behavior, but be aware it may not always make sense. Elasticsearch's current behavior is also another resource for how OpenSearch should behave.
    * Let's be pragmatic about this – our foremost goal is the migration of OpenSearch's frontend behavior. We can aim for some Elasticsearch feature parity, but only _within reason_.
* Often there is a PR in the Elasticsearch backend data source migration [https://github.com/grafana/grafana/issues/54011](https://github.com/grafana/grafana/issues/54011) which is 75% of what OpenSearch will need
* Migrate or implement: i.e. make any changes to code in backend
* Use the dashboard created above to compare visualization results (data frame comparison)
    * Can also inspect data frame on a more detailed basis

iwysiu:

This make PPL queries with the `table` format run in `Explore` execute on the backend.

You can use a PPL query from the "PPL Table" section [here](https://clouddatasources.grafana.net/goto/vMOq6OmIg?orgId=1) to test it with the external dev.

Fixes #201 

Similar to https://github.com/grafana/opensearch-datasource/pull/259 , in order to test the ad-hoc filters you have to enable ppl queries to be sent from the dashboard to the backend (remove the check for `CoreApp.Explore` in `datasource.ts` at line 548)

Also: my query in `ppl_table.query_input.json` isn't actually the query that produces the response I'm using (i narrowed it down a bit to be smaller) since it seemed like that was allowed. If they should match I can make them.

### DIFF
--- a/pkg/opensearch/ppl_response_parser.go
+++ b/pkg/opensearch/ppl_response_parser.go
@@ -127,8 +127,10 @@ func (rp *pplResponseParser) parsePPLResponse(queryRes *backend.DataResponse, co
 	if frame.Meta == nil {
 		frame.Meta = &data.FrameMeta{}
 	}
-	if isLogsQuery {
-		frame.Meta.PreferredVisualization = data.VisTypeLogs
+
+	frame.Meta.PreferredVisualization = data.VisTypeLogs
+	if !isLogsQuery {
+		frame.Meta.PreferredVisualization = data.VisTypeTable
 	}
 
 	queryRes.Frames = append(queryRes.Frames, frame)
--- a/pkg/opensearch/ppl_response_parser.go
+++ b/pkg/opensearch/ppl_response_parser.go
@@ -105,6 +105,7 @@ func (rp *pplResponseParser) parsePPLResponse(queryRes *backend.DataResponse, co
 
 		doc = flatten(doc, maxFlattenDepth)
 		for key := range doc {
+			// Do not add _source field (besides logs) as we are showing each _source field in table instead
 			if !isLogsQuery && key == "_source" {
 				continue
 			}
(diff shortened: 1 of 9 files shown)


pkg/opensearch/ppl_response_parser.go
```diff
@@ -100,14 +116,22 @@ func (rp *pplResponseParser) parseLogs(queryRes *backend.DataResponse, configure
...
 	fields := processDocsToDataFrameFields(docs, sortedPropNames)
 
 	frame := data.NewFrame("", fields...)
 	if frame.Meta == nil {
 		frame.Meta = &data.FrameMeta{}
 	}
+
 	frame.Meta.PreferredVisualization = data.VisTypeLogs
+	if !isLogsQuery {
+		frame.Meta.PreferredVisualization = data.VisTypeTable
```
iwysiu, line 133: technically this isn't in the frontend code (it doesn't set it) but it made sense as an improvement to me

pkg/opensearch/ppl_response_parser.go
```diff
@@ -86,12 +99,16 @@ func (rp *pplResponseParser) parseLogs(queryRes *backend.DataResponse, configure
...
 		}
 
-		if configuredFields.LogLevelField != "" {
+		if isLogsQuery && configuredFields.LogLevelField != "" {
 			doc["level"] = doc[configuredFields.LogLevelField]
 		}
 
 		doc = flatten(doc, maxFlattenDepth)
 		for key := range doc {
+			// Do not add _source field (besides logs) as we are showing each _source field in table instead
```
iwysiu, line 108: I copied this comment from the frontend code, honestly it's not that clear to me, I can update it if someone has a better understanding of what it means.

kevinwcyu: my understanding is that `_source` is the original payload that elasticsearch/opensearch indexes.

e.g.

```json
{
  "src": "CA",
  "dst": "US",
  "start_date": "2023-10-25",
  "end_date": "2023-10-26",
  "geo": {
    "lat": 49.00000,
    "lon": 10.00000
  }
}
```

when you use the logs panel to display a logs query, i think it treats the `_source` field as the log message field (the field that is the original payload that the data is derived from) unless you specify another field.

for table formatted queries, you probably wouldn't want a column that just displays the json of the original payload since each of the fields in the original payload is already its own column.

kevinwcyu (approved):

Issue #287, opened 2023-10-17
Migrate trace list response processing


fridgepoet:
"Trace list" has already been implemented in the frontend.
In the backend, the query to OpenSearch has been implemented in https://github.com/grafana/opensearch-datasource/pull/284.

For this task, the response from OpenSearch should be processed into a response for Grafana. 
* This document has instructions on how to run the “opensearch dashboards” create traces locally with test data:  https://docs.google.com/document/d/1l8HYKvCddF9jUdkAV0TkHU6fsWFs8T1s-gLWbaGLtog/edit
* Then you can use the current frontend implementation to "1. View all traces:" in the description here: https://github.com/grafana/opensearch-datasource#traces-support . This is a "trace list"

"Trace list" has been described in: https://docs.google.com/document/d/1XNGvwURUkMHCKi4kjFOZDIBhIclnmO-d-_CsD8iTqVA/edit and https://docs.google.com/document/d/1l8HYKvCddF9jUdkAV0TkHU6fsWFs8T1s-gLWbaGLtog/edit

fridgepoet:
Sure I'll add to the description. Please note this is a task of the epic https://github.com/grafana/opensearch-datasource/issues/223

PR #305, opened 2023-11-15
Traces bug on Explore

sarahzinger:
While working on https://github.com/grafana/opensearch-datasource/issues/287 I noticed that when I am in the explore tab and make a traces request, the request goes through the backend flow, and when I make a request through the Query Editor it goes through the frontend flow. 

But we haven't actually done the work to support the backend flow for traces yet (that's part of what that ticket is about), so if users attempt to make traces queries in Explore mode they are not going to get data back. 

Until this work is done I think we should add a case to keep just the frontend feature flow going.

### DIFF
--- a/src/datasource.ts
+++ b/src/datasource.ts
@@ -541,7 +541,7 @@ export class OpenSearchDatasource extends DataSourceWithBackend<OpenSearchQuery,
               metric.type === 'raw_document' ||
               (request.app === CoreApp.Explore &&
                 target.queryType === QueryType.Lucene &&
-                target.luceneQueryType !== LuceneQueryType.Traces)
+                target.luceneQueryType === LuceneQueryType.Metric)
           ) ||
           (request.app === CoreApp.Explore &&
             target.queryType === QueryType.PPL &&
--- a/src/datasource.ts
+++ b/src/datasource.ts
@@ -541,7 +541,7 @@ export class OpenSearchDatasource extends DataSourceWithBackend<OpenSearchQuery,
               metric.type === 'raw_document' ||
               (request.app === CoreApp.Explore &&
                 target.queryType === QueryType.Lucene &&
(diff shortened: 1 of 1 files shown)


src/datasource.ts
```diff
@@ -538,12 +534,14 @@ export class OpenSearchDatasource extends DataSourceWithBackend<OpenSearchQuery,
...
+        (target) =>
           target.metrics?.every(
-            metric =>
+            (metric) =>
               metric.type === 'raw_data' ||
               metric.type === 'raw_document' ||
-              (request.app === CoreApp.Explore && target.queryType === QueryType.Lucene)
+              (request.app === CoreApp.Explore &&
+                target.queryType === QueryType.Lucene &&
+                target.luceneQueryType !== LuceneQueryType.Traces)
```
sarahzinger, line 544: sorry the git diff here is a lot because my linter ran or something, but this is the only line I really intended to change.

kevinwcyu: can it be `target.luceneQueryType === LuceneQueryType.Metric` instead to make it explicit what is being sent to the backend.

fridgepoet:
WOW sorry about that

fridgepoet (approved):

idastambuk (approved):

src/datasource.ts
sarahzinger: ooh nice call

sarahzinger: lol so I tried this and broke a bunch of tests. Not entirely sure if this reflecting issues with tests or code (I'm guessing the second) but I'm going to just merge as is for now, hopefully this code is pretty temporary anyway

sarahzinger:
> WOW sorry about that

No worries! I think this was actually never released so it's kind of like it never happened :)

PR #310, opened 2023-11-20
All trace list requests go through backend

### LINKED ISSUE #287: Migrate trace list response processing

sarahzinger:
Before Merging:
- [x] merge https://github.com/grafana/opensearch-datasource/pull/309 and then edit this pr to be based off of main

Closes #287

Adds support for the Trace List view showing all traces through a backend flow, but parsing the response and enabling it in js


sarahzinger:
I think this pr is failing because of code coverage which is interesting because I thought we removed it? I wonder if we run it on branches that are not based on main or something like that?

kevinwcyu:
@sarahzinger test failure is related to https://github.com/grafana/oss-plugin-partnerships/issues/697, unfortunately, i think restarting the test is the only solution for now.

pkg/opensearch/response_parser.go
```diff
@@ -97,21 +97,26 @@ func (rp *responseParser) parseResponse() (*backend.QueryDataResponse, error) {
...
 		}
 
-		// trace span condition
-		// trace queries are sent from the FE with a metrics field, returning early so the switch doesn't overwrite the response from the traces query
-		if target.luceneQueryType == luceneQueryTypeTraces {
-			queryRes = processTraceSpansResponse(res, queryRes)
-			result.Responses[target.RefID] = queryRes
-			continue
+		var queryType string
+		if target.luceneQueryType == "Traces" {
```
iwysiu, line 101: we should still use the `luceneQueryTypeTraces` const here

pkg/opensearch/response_parser.go
```diff
@@ -265,6 +271,56 @@ func processTraceSpansResponse(res *es.SearchResponse, queryRes backend.DataResp
...
+	// but es.SearchResponse is deliberately not typed as in other query cases it can be much more open ended
+	rawTraces := res.Aggregations["traces"].(map[string]interface{})["buckets"].([]interface{})
+
+	// get values from raw traces response
+	traceIds := []string{}
+	traceGroups := []string{}
+	traceLatencies := []float64{}
+	traceErrorCounts := []float64{}
+	traceLastUpdated := []time.Time{}
+	for _, trace := range rawTraces {
```
iwysiu, line 286: Nit: I think this would be a little clearer if `t` and `trace` swapped names. Or if `t` had a more descriptive name

pkg/opensearch/response_parser.go
```diff
@@ -265,6 +271,56 @@ func processTraceSpansResponse(res *es.SearchResponse, queryRes backend.DataResp
...
+		t := trace.(map[string]interface{})
+
+		traceIds = append(traceIds, t["key"].(string))
+		traceGroups = append(traceGroups, t["trace_group"].(map[string]interface{})["buckets"].([]interface{})[0].(map[string]interface{})["key"].(string))
```
iwysiu, line 290: Did you end up trying the `flatten` helper?

iwysiu (commented):
I was thinking about this this weekend, and I realized another possible solution is writing a custom unmarshaller for es.SearchResponse and adding a field for a more defined struct to store the trace list response if we have one. (We can try writing one together if you want)

Also, should this have a [shapshot test](https://github.com/grafana/opensearch-datasource/tree/main/pkg/opensearch/snapshot_tests)?

pkg/opensearch/response_parser.go
sarahzinger: I played around with it a bit, but I don't think it helps in this case but maybe I'm misunderstanding?

sarahzinger:
> I was thinking about this this weekend, and I realized another possible solution is writing a custom unmarshaller for es.SearchResponse and adding a field for a more defined struct to store the trace list response if we have one. (We can try writing one together if you want)
> 

Sounds neat! Maybe we can do it in a future team time in a separate pr if that sounds good? Or if you think that should be a blocker to this let me know and we can prioritize it

> Also, should this have a [shapshot test](https://github.com/grafana/opensearch-datasource/tree/main/pkg/opensearch/snapshot_tests)?

Yes! Great catch I'll do that now 😄

iwysiu (approved):
> Sounds neat! Maybe we can do it in a future team time in a separate pr if that sounds good? Or if you think that should be a blocker to this let me know and we can prioritize it

Considering that this is a blocker for other work, I'm ok with merging this for now.

PR #315, opened 2023-12-01
Upgrade grafana dependencies

idastambuk:
- upgrades all Grafana dependencies to 10.2.0
- upgrades plugin-tools' create plugin, hence the many changes in configuration files
- fixes tests to use render() from react-testing-library cause there were some type errors. We should probably remove enzyme at some point unless it's needed
- There are some deprecation errors in the build log ```'datasourceRequest' is deprecated. Use the fetch function instead```, for example. This should probably be fixed in a refactor, but I didn't want to make any additional changes
- there were some type errors with min_doc_count in useDescription.ts, so I added a parseInt(). Checked in Elastic, this is what they did as well: https://github.com/grafana/grafana/blob/main/public/app/plugins/datasource/elasticsearch/components/QueryEditor/BucketAggregationsEditor/SettingsEditor/useDescription.ts#L74

Fixes https://github.com/grafana/oss-plugin-partnerships/issues/742


.config/.cprc.json
```diff
@@ -0,0 +1,3 @@
+{
```
kevinwcyu, line 1: What exactly is this file for? I tried googling `.cprc.json` but couldn't find what it's for

package.json
```diff
@@ -15,22 +15,23 @@
     "test:ci": "jest --passWithNoTests --maxWorkers 4",
     "test:coverage": "jest --coverage",
     "test:coverage:changes": "jest --coverage --changedSince=origin/main",
-    "typecheck": "tsc --noEmit"
+    "typecheck": "tsc --noEmit",
+    "e2e": "yarn exec cypress install && yarn exec grafana-e2e run",
+    "e2e:update": "yarn exec cypress install && yarn exec grafana-e2e run --update-screenshots"
```
kevinwcyu, line 20: These 2 can be removed for now since we don't have e2e tests here yet, right?

kevinwcyu (approved):

.config/.cprc.json
idastambuk: It's a file generated by create-plugin, I think it might be how plugin tools trace the version of create-plugin for upgrades etc.

PR #319, opened 2023-12-18
Refactor tests to remove Enzyme and use react-testing-library

idastambuk:
This refactors tests a bit to remove Enzyme dependency and use react-testing-library everywhere. Since we don't need enzyme-specific stuff, it's one fewer dependency to worry about when upgrading and dealing with vulnerabilities. 

Fixes #


src/components/QueryEditor/QueryTypeEditor/index.tsx
```diff
@@ -30,13 +30,15 @@ export const QueryTypeEditor = ({ value }: Props) => {
...
   return (
-    <Segment
-      className={segmentStyles}
-      options={getQueryTypeOptions(datasource.getSupportedQueryTypes())}
-      onChange={(e) => {
-        isValidQueryType(e.value) && dispatch(changeQueryType(e.value!));
-      }}
-      value={toOption(value)}
-    />
+    <div data-testid="query-type-wrapper">
```
idastambuk, line 33: Had to add the div wrapper as this is the only way to test Segment with selectEvent library. It doesn't change the UI at all

src/components/QueryEditor/PPLFormatEditor/index.test.tsx
```diff
@@ -1,39 +1,51 @@
...
-    wrapper
-      .find(OpenCloseButton)
-      .props()
-      .onClick();
-    expect(wrapper.find(HelpMessage).length).toBe(1);
+  it('should show help message on click', async () => {
+    setup();
+    await waitFor(async () => {
+      const button = screen.getByText('Show help');
+      await userEvent.click(button);
```
sarahzinger, line 47: WaitFor is a tricky beast, I'd recommend checking out https://kentcdodds.com/blog/common-mistakes-with-react-testing-library there are a few instances in this pr that I think might be worth taking a look at how we're using it.

For example in this case, if I understand correctly waitFor will actually call multiple times, so in this case you're basically clicking the button multiple times. 

Might be good to see is it possible to not use waitfor here? If not figure out why, what exactly are we waiting for? What in the dom can we use as a marker to say "hey the thing we are waiting for is done". Do we need to wait for all of these things for example? Or could we wait for the button to show the correct text and then continue? 

Could we use find instead of waitfor, etc.

src/components/QueryEditor/QueryTypeEditor/index.tsx
sarahzinger: seems like a bug in Segment! Feels like we should be able to pass it a testid, could be worth opening a ticket?

sarahzinger (approved):
Nice work!!!

src/components/QueryEditor/PPLFormatEditor/index.test.tsx
idastambuk: Didn't know this about the waitFor callback! The reason I'm waiting for the setup is that there is some useContext stuff that needs to happen for the component to render, but I can certainly use find, that would also make a lot of sense.

idastambuk:
@sarahzinger So another Christmas miracle - went to see where the testId ends up in Segment, but now the test works 🤷🏻‍♀️🎁

PR #321, opened 2024-01-13
[BUG] Support time field with nanoseconds

Garfield96:
The time field for traces can have nanosecond precision. If nanoseconds were used, the following error was shown:
image
This PR fixes this issue by adapting the timeformat to allow for arbitrary subsecond precision up to nanoseconds. 

N/A

It would be great if this change could be released as a patch release, since I can't use the plugin without this change. Thanks

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -12,7 +12,7 @@ import (
 )
 
 const (
-	TimeFormat = "2006-01-02T15:04:05.000000Z"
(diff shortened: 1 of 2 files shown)


idastambuk (approved):

idastambuk:
Hi @Garfield96 thanks a lot for the contribution! This will be released with https://github.com/grafana/opensearch-datasource/pull/322 as a patch

Garfield96:
Hi @idastambuk,
thanks for the fast review and the release of the new version. The issue is now resolved.

PR #324, opened 2024-01-30
Fix: Move "Build a release" in CONTRIBUTING.md out of TLS section

fridgepoet:
Probably while writing the TLS section I accidentally moved "Build a release" to the wrong section


njvrzm (approved):
👍🏻

idastambuk (approved):

PR #325, opened 2024-02-01
Backend migration: logs queries should go through frontend in Explore

idastambuk:
We received a [bug report ](https://github.com/grafana/support-escalations/issues/9187)that data links don't work in explore. 
This would normally be an easy fix: we would do something [like this in ElasticSearch](https://github.com/grafana/grafana/blob/62c8584443c1163f936d54a7fd94067ec75abe2c/public/app/plugins/datasource/elasticsearch/datasource.ts#L671). However, this doesn't seem to work with plugin sandboxing. Until we figure out how to pass the new response, we should revert logs to go through the frontend again. 

Fixes #

### DIFF
--- a/src/datasource.test.ts
+++ b/src/datasource.test.ts
@@ -1236,30 +1236,30 @@ describe('OpenSearchDatasource', function (this: any) {
       expect(mockedSuperQuery).not.toHaveBeenCalled();
     });
 
-    it('should send logs queries in Explore', () => {
-      const mockedSuperQuery = jest
-        .spyOn(DataSourceWithBackend.prototype, 'query')
-        .mockImplementation((request: DataQueryRequest<OpenSearchQuery>) => of());
-      const logsQuery: OpenSearchQuery = {
-        refId: 'A',
-        metrics: [{ type: 'logs', id: '1' }],
-        query: 'foo="bar"',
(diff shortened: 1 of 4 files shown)


src/datasource.test.ts
```diff
@@ -1535,6 +1535,31 @@ describe('OpenSearchDatasource', function (this: any) {
       expect(mockedSuperQuery).not.toHaveBeenCalled();
       expect(datasourceRequestMock).toHaveBeenCalled();
     });
+    it('should not send logs queries in Explore ot backend', () => {
```
kevinwcyu, line 1538:
```suggestion
    it('should not send logs queries in Explore to backend', () => {
```

src/datasource.test.ts
```diff
@@ -1679,6 +1704,31 @@ describe('OpenSearchDatasource', function (this: any) {
       expect(mockedSuperQuery).not.toHaveBeenCalled();
       expect(datasourceRequestMock).toHaveBeenCalled();
     });
+     it('should send PPL logs format queries in Explore', () => {
```
kevinwcyu, line 1707:
```suggestion
    it('should not send PPL logs format queries in Explore to the backend', () => {
```

kevinwcyu (approved):

PR #326, opened 2024-02-02
Backend Migration: Add data links to responses from the backend

idastambuk:

Adds response processing for data links in the frontend to mirror ElasticSearch: https://github.com/grafana/grafana/blob/main/public/app/plugins/datasource/elasticsearch/datasource.ts#L671

We could probably do this on the backend as well, but due to this being reported in an escalation, this is quicker.

Fixes https://github.com/grafana/support-escalations/issues/9187

To test, go to datasource config for the provisioning OpenSearch datasource. Then add a data link with name: host and select a datasource. After this, logs data frames in explore should have links at the bottom of each log entry.

### DIFF
--- a/src/datasource.test.ts
+++ b/src/datasource.test.ts
@@ -56,7 +56,7 @@ jest.mock('@grafana/runtime', () => ({
   getDataSourceSrv: () => {
     return {
       getInstanceSettings: () => {
-        return { name: 'elastic25' };
+        return { name: 'OSds' };
       },
     };
   },
@@ -2152,7 +2152,7 @@ describe('OpenSearchDatasource', function (this: any) {
         url: '',
         internal: {
           query: { query: 'query' },
-          datasourceName: 'elastic25',
+          datasourceName: 'OSds',
           datasourceUid: 'dsUid',
         },
       });
@@ -2198,7 +2198,7 @@ describe('enhanceDataFrame', () => {
       url: '',
       internal: {
         query: { query: 'query' },
-        datasourceName: 'elastic25',
+        datasourceName: 'OSds',
         datasourceUid: 'dsUid',
       },
     });
--- a/src/datasource.test.ts
+++ b/src/datasource.test.ts
@@ -3,6 +3,7 @@ import {
   CoreApp,
   DataFrame,
   DataQueryRequest,
+  DataQueryResponse,
   DataSourceInstanceSettings,
   DateTime,
   dateTime,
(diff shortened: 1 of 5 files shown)


src/types.ts
```diff
@@ -85,6 +85,7 @@ export type DataLinkConfig = {
   field: string;
   url: string;
   datasourceUid?: string;
+  urlDisplayLabel?: string;
```
kevinwcyu, line 88: We can remove this. It's not available as an option in the config editor and the current frontend implementation uses an empty string for the data link `title` anyway.

kevinwcyu (approved):

src/datasource.ts
```diff
@@ -557,13 +557,9 @@ export class OpenSearchDatasource extends DataSourceWithBackend<OpenSearchQuery,
...
       const adHocAndInterpolatedRequest = { ...request, targets: queriesWithAdHocAndInterpolatedVariables };
       return super.query(adHocAndInterpolatedRequest).pipe(
-        tap({
-          next: (response) => {
-            trackQuery(response, targetsWithInterpolatedVariables, adHocAndInterpolatedRequest.app);
-          },
-          error: (error) => {
-            trackQuery({ error, data: [] }, targetsWithInterpolatedVariables, adHocAndInterpolatedRequest.app);
-          },
+        tap((response) => trackQuery(response, targetsWithInterpolatedVariables, adHocAndInterpolatedRequest.app)),
```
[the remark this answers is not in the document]
idastambuk: Good catch, we shouldn't indeed lose it, I put it back

src/utils.ts
```diff
@@ -129,3 +130,55 @@ export const createEmptyDataFrame = (
 
   return series;
 };
+
+export function enhanceDataFramesWithDataLinks(newResponse: DataQueryResponse, dataLinks: DataLinkConfig[]) {
```
sarahzinger, line 134: Can we add a unit test for this one?

src/datasource.test.ts
```diff
@@ -2094,6 +2095,69 @@ describe('OpenSearchDatasource', function (this: any) {
       });
     });
   });
+  describe('Data links', () => {
+    it('should add links to dataframe for logs queries in the backend flow', async () => {
```
sarahzinger, line 2099: maybe remove "in the backend flow"

sarahzinger (commented):
I never noticed this feature, it's neat, thanks for fixing!

PR #344, opened 2024-03-07
Allow to use script in query variable

### LINKED ISSUE #343: Support for script field in query variables
I need to have a query variable that aggregate on boolean. However, the boolean aggregation return the number 1 or 0 in the bucket key. This is a problem when Grafana uses this value in the main query because it uses 1 or 0 but OpenSearch query expect "true" or "false" to match a document attribute with boolean type.
https://opensearch.org/docs/latest/field-types/supported-field-types/boolean/#boolean-values-in-aggregations-and-scripts

According to the documentation, we can only use "field" in the query variable aggregation:
https://github.com/grafana/opensearch-datasource/tree/main?tab=readme-ov-file#query-variable

I think we should allow _also_ for the "script" field to enable advanced logic, for example

```json
{"find": "terms", "script":"doc['isActive'].value.toString()", "query": "hubName:$hub_name"}
```

from my early analysis, it would be a matter of adding the "script" mapping here to allow using the field
https://github.com/grafana/opensearch-datasource/blob/main/src/QueryBuilder.ts#L390

and here to allow the replacement of variables by Grafana
https://github.com/grafana/opensearch-datasource/blob/main/src/datasource.ts#L957

no alternative

loru88:
Allow to use script in query variable, so that one can use field for simple use case and painless script for advanced use case

Fixes #343

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,9 @@ The OpenSearch data source supports two types of queries you can use in the _Que
 | `{"find": "fields", "type": "keyword"}`                              | Returns a list of field names with the index type `keyword`.                                                                                                          |
 | `{"find": "terms", "field": "@hostname", "size": 1000}`              | Returns a list of values for a field using term aggregation. Query will use current dashboard time range as time range for query.                                     |
 | `{"find": "terms", "field": "@hostname", "query": '<lucene query>'}` | Returns a list of values for a field using term aggregation and a specified lucene query filter. Query will use current dashboard time range as time range for query. |
+| `{"find": "terms", "script": "if( doc['@hostname'].value == 'something' ) { return null; } else { return doc['@hostname']}"}`, "query": '<lucene query>'}` | Same as above but you can use the script API instead |
+
+If the query is multi-field with both a text and keyword type, use "field":"fieldname.keyword" (sometimes fieldname.raw) to specify the keyword field in your query.
 
 There is a default size limit of 500 on terms queries. Set the size property in your query to set a custom limit.
 You can use other variables inside the query. Example query definition for a variable named `$host`.
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ The OpenSearch data source supports two types of queries you can use in the _Que
 | `{"find": "fields", "type": "keyword"}`                              | Returns a list of field names with the index type `keyword`.                                                                                                          |
 | `{"find": "terms", "field": "@hostname", "size": 1000}`              | Returns a list of values for a field using term aggregation. Query will use current dashboard time range as time range for query.                                     |
 | `{"find": "terms", "field": "@hostname", "query": '<lucene query>'}` | Returns a list of values for a field using term aggregation and a specified lucene query filter. Query will use current dashboard time range as time range for query. |
-| `{"find": "terms", "script": "if( doc['@hostname'].value == 'something' ) { return null; } else { return doc['@hostname']}"}`, "query": '<lucene query>'}` | Same as above but you can use the script API instead |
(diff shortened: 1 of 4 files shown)


iwysiu (commented):
Hi @loru88 ! Thanks for the contribution, it sounds like a good improvement. Have you tried it successfully with the query you were trying to run? I added a small change request, and it also might be nice to add this to the documentation in the readme.

src/datasource.ts
```diff
@@ -954,7 +954,12 @@ export class OpenSearchDatasource extends DataSourceWithBackend<OpenSearchQuery,
       }
 
       if (parsedQuery.find === 'terms') {
-        parsedQuery.field = getTemplateSrv().replace(parsedQuery.field, {}, 'lucene');
+        if (parsedQuery.field) {
+          parsedQuery.field = getTemplateSrv().replace(parsedQuery.field, {}, 'lucene');
+        }
+        if (parsedQuery.script) {
+          parsedQuery.script = getTemplateSrv().replace(parsedQuery.script);
```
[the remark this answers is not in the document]
loru88: Hello, yes it works as expected.

I thought not to put "lucene" formatting because that field is not actually a Lucene query, is a Painless scripting language field: https://opensearch.org/docs/latest/api-reference/script-apis/exec-script/

loru88:
> Sounds good, just as a last change could you add the new format to the table of allowed query variables in the [README.md](https://github.com/grafana/opensearch-datasource/blob/4ea4e370a3d6150be8800169aded9e479818259b/README.md?plain=1#L128)?

Thank you for reminding me, I forgot about that piece.
I also have added a clarification about using `.keyword`

README.md
```diff
@@ -130,6 +130,9 @@ The OpenSearch data source supports two types of queries you can use in the _Que
 | `{"find": "fields", "type": "keyword"}`                              | Returns a list of field names with the index type `keyword`.                                                                                                          |
 | `{"find": "terms", "field": "@hostname", "size": 1000}`              | Returns a list of values for a field using term aggregation. Query will use current dashboard time range as time range for query.                                     |
 | `{"find": "terms", "field": "@hostname", "query": '<lucene query>'}` | Returns a list of values for a field using term aggregation and a specified lucene query filter. Query will use current dashboard time range as time range for query. |
+| `{"find": "terms", "script": "if( doc['@hostname'].value == 'something' ) { return null; } else { return doc['@hostname']}"}`, "query": '<lucene query>'}` | Same as above but you can use the script API instead |
```
iwysiu, line 133: Just fixing the formatting and making the language match
```suggestion
| `{"find": "terms", "script": "if( doc['@hostname'].value == 'something' ) { return null; } else { return doc['@hostname']}", "query": '<lucene query>'}` | Returns a list of values using term aggregation, the script API, and a specified lucene query filter. Query will use current dashboard time range as time range for query. |
```

README.md
```diff
@@ -130,6 +130,9 @@ The OpenSearch data source supports two types of queries you can use in the _Que
...
 | `{"find": "terms", "field": "@hostname", "size": 1000}`              | Returns a list of values for a field using term aggregation. Query will use current dashboard time range as time range for query.                                     |
 | `{"find": "terms", "field": "@hostname", "query": '<lucene query>'}` | Returns a list of values for a field using term aggregation and a specified lucene query filter. Query will use current dashboard time range as time range for query. |
+| `{"find": "terms", "script": "if( doc['@hostname'].value == 'something' ) { return null; } else { return doc['@hostname']}", "query": '<lucene query>'}` | Returns a list of values using term aggregation, the script API, and a specified lucene query filter. Query will use current dashboard time range as time range for query. |
+
+If the query is multi-field with both a text and keyword type, use "field":"fieldname.keyword" (sometimes fieldname.raw) to specify the keyword field in your query.
```
iwysiu, line 135:
```suggestion
If the query is multi-field with both a text and keyword type, use `"field":"fieldname.keyword"` (sometimes `fieldname.raw`) to specify the keyword field in your query.
```

iwysiu (approved):
Approved! I'll get this merged.

Issue #330, opened 2024-02-14
Backend migration - Forward OAuth authentication stops working after 2.13


mimek:
**What happened**:
After upgrading to version 2.13.0 of the plugin, query from "Explore view" to OpenSearch stopped working. We used "Forward OAuth" option in datasource configuration to authenticate to OpenSearch.

Dashboards visualisations which uses OpenSearch datasource still works fine.

**What you expected to happen**:
Consistent behaviour of querying through "Explore view" and Dashboard. When you configure datasource and test, connectivity test is saying that everything works fine, but if you choose "Explore view" then queries are not working (in debug logs there's visible 401 error while connecting to OpenSearch). After adding configured datasource to dashboard it's working again.

**How to reproduce it (as minimally and precisely as possible)**:
Configure opensearch datasource with "Forward OAuth" option enable (Opensearch must be also configured to support this authentication mechanism) and go to "Explore view" and execute some query.
![image](https://github.com/grafana/opensearch-datasource/assets/6304986/e487c288-0219-4dbf-9475-6a958288652e)

Then try to add same OpenSearch datasource to dashboard and see that's working there.
![image](https://github.com/grafana/opensearch-datasource/assets/6304986/aa6a1102-1eca-4c22-bdda-d82b98dcf864)

**Anything else we need to know?**:

**Environment**:
- Grafana version: 9.5.15, 10.3.1
- OpenSearch version: 2.11.1
- Plugin version: 2.13

iwysiu:
Hi @mimek ! Thanks for reporting this issue.

For more detail on what's happening, we've done some refactoring of the Opensearch plugin, and currently the Explore and Dashboard queries are running through different paths in case of bugs like this, I'll move it to our backlog to be fixed.

Investigation notes for the person who fixes this:
It looks like (unlike the frontend path) the `httpclient.Provider` we're using for backend queries doesn't automatically forward oauth tokens. We should set `ForwardHTTPHeaders` on the http options to true and it should be passed. See the [developer notes](https://grafana.com/developers/plugin-tools/create-a-plugin/extend-a-plugin/add-authentication-for-data-source-plugins#forward-oauth-identity-for-the-logged-in-user)

mimek:
Hi @iwysiu for explanations and confirming that it should be fixed!

idastambuk:
Hi @mimek we just released v2.14.5 of the plugin that should fix this issue. Can you confirm that your explore queries work with Forward Auth with the new version? Thanks!

PR #345, opened 2024-03-12
Bugfix: Forward http headers to enable OAuth for backend queries

idastambuk:

After upgrading to version 2.13.0 of the plugin, query from "Explore view" to OpenSearch stopped working. The user used "Forward OAuth" option in datasource configuration to authenticate to OpenSearch.
This passes on the headers to be used to authenticate down stream. 

From @iwysiu's investigation: 
```
Investigation notes for the person who fixes this:
It looks like (unlike the frontend path) the httpclient.Provider we're using for backend queries doesn't automatically forward oauth tokens. We should set ForwardHTTPHeaders on the http options to true and it should be passed. See the [developer notes](https://grafana.com/developers/plugin-tools/create-a-plugin/extend-a-plugin/add-authentication-for-data-source-plugins#forward-oauth-identity-for-the-logged-in-user)
```
Fixes https://github.com/grafana/opensearch-datasource/issues/330

There doesn't seem to be a simple way to unit test this without a refactor to mock out the client, so I decided not to add one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ All notable changes to this project will be documented in this file.
 
 ## 2.14.3
 
+- Bugfix: Forward http headers to enable OAuth for backend queries [#345](https://github.com/grafana/opensearch-datasource/pull/345)
+
+## 2.14.3
+
 - Update grafana-aws-sdk to 0.22.0 in [#323](https://github.com/grafana/opensearch-datasource/pull/323)
 
 ## 2.14.2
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,11 +2,19 @@
...
+- Allow to use script in query variable by @loru88 in [#344](https://github.com/grafana/opensearch-datasource/pull/344)
+- Chore: run go mod tidy #338 [#338](https://github.com/grafana/opensearch-datasource/pull/338)
+- Chore: adds basic description and link to github by @sympatheticmoose in [#337](https://github.com/grafana/opensearch-datasource/pull/337)
+
 ## 2.14.4
 
 - Fix: Move "Build a release" in CONTRIBUTING.md out of TLS section in [#324](https://github.com/grafana/opensearch-datasource/pull/324)
 - Backend Migration: Add data links to responses from the backend in [#326](https://github.com/grafana/opensearch-datasource/pull/326)
 
+
```
iwysiu, line 17:
```suggestion
```

iwysiu (approved):

PR #347, opened 2024-03-15
Add keywords

kevinwcyu:
Add keywords to improve search in the `/connections/add-new-connection` page of Grafana

### DIFF
--- a/src/plugin.json
+++ b/src/plugin.json
@@ -16,7 +16,7 @@
       "name": "Grafana Labs",
       "url": "https://grafana.com"
     },
-    "keywords": ["elasticsearch", "opensearch"],
+    "keywords": ["elasticsearch", "opensearch", "aws", "cloud provider", "database", "logs", "nosql"],
     "logos": {
       "small": "img/logo.svg",
       "large": "img/logo.svg"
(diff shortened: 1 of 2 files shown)


src/plugin.json
```diff
@@ -16,7 +16,7 @@
       "name": "Grafana Labs",
       "url": "https://grafana.com"
     },
-    "keywords": ["elasticsearch", "opensearch"],
+    "keywords": ["elasticsearch", "opensearch", "aws", "cloud provider", "database", "logs", "nosql"],
```
iwysiu, line 19: Should this include traces?

sarahzinger (approved):
I second the suggestion for traces... I'm also not 100% on whether nosql would be considered appropriate here or not. I think it probably is appropriate just maybe not the most common use case.

kevinwcyu:
> I second the suggestion for traces... I'm also not 100% on whether nosql would be considered appropriate here or not. I think it probably is appropriate just maybe not the most common use case.

I added it because I saw that Elasticsearch included it as well and it's also listed as a type of NoSQL DB here https://aws.amazon.com/nosql

iwysiu (approved):

Issue #334, opened 2024-02-20
FE: Add "Node Graph" switch to the Query Editor


idastambuk:
We should add a "Node Graph" switch to the Query Editor when query type is "traces" and send a nodeGraph:boolean flag int he query object to the BE

PR #350, opened 2024-03-18
Add node graph toggle

sarahzinger:
Adds a node graph toggle to the query editor, doesn't do much yet except send over the boolean to the backend. 
Screenshot 2024-03-18 at 3 28 57 PM

Fixes https://github.com/grafana/opensearch-datasource/issues/334

I've done this behind a feature toggle, but did not put the feature toggle into main grafana. We could add it to main grafana if we wanted to but I am not sure it's worth it in this case, it just gives us better typing and things like that. I think it might be enough to just sort of ad-hoc add it as I've done here so that we're able to work on this in multiple prs and merge/release before the full feature is done.


idastambuk (approved):

idastambuk:
I like the solution with the temporary toggle!

PR #355, opened 2024-04-09
Traces: Direct all trace queries to the BE and rename Node graph to Service map

idastambuk:
I think it makes sense for us to redirect all trace queries to the backend before we release the Service Map feature. It's a BE feature and if we don't redirect all trace queries to the BE, we won't have feature parity in Dashboards vs. Explore

I also renamed node graph to Service Map, since it looks like 'service map' is used elswhere (X-ray, Elastic, Tempo) for the feature, and Node Graph is used for the Grafana visualization. 

### DIFF
--- a/pkg/opensearch/response_parser_test.go
+++ b/pkg/opensearch/response_parser_test.go
@@ -2659,3 +2659,249 @@ func TestProcessTraceListResponse(t *testing.T) {
 	assert.Equal(t, "Last Updated", lastUpdated.Name)
 	assert.Equal(t, "time.Time", lastUpdated.Type().ItemTypeString())
 }
+
+
+func TestProcessSpansResponse_withMultipleSpansQueries(t *testing.T) {
+	targets := map[string]string{
+		"A": `{
+			"timeField": "@timestamp",
+			"metrics": [{ "type": "count", "id": "1" }],
+			"query": "traceId:test",
+			"luceneQueryType": "Traces"
+			}`,
+		"B": `{
+			"timeField": "@timestamp",
+			"metrics": [{ "type": "count", "id": "1" }],
+			"query": "traceId:test123",
+			"luceneQueryType": "Traces"
+			}`,
+	}
+
+	response := `
+	{
+		"responses": [
+			{
+				"hits": {
+					"hits": [
+						{
+							"_source": {
+								"traceId": "000000000000000047ed3a25a7dba0cd",
+								"droppedLinksCount": 0,
+								"kind": "SPAN_KIND_SERVER",
+								"droppedEventsCount": 0,
+								"traceGroupFields": {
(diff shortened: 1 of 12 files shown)


src/datasource.ts
```diff
@@ -544,9 +544,7 @@ export class OpenSearchDatasource extends DataSourceWithBackend<OpenSearchQuery,
           (request.app === CoreApp.Explore &&
             target.queryType === QueryType.PPL &&
             (target.format === 'logs' || target.format === 'table')) ||
-          (request.app === CoreApp.Explore &&
-            target.luceneQueryType === LuceneQueryType.Traces &&
-            getTraceIdFromLuceneQueryString(target.query ?? ''))
```
iwysiu, line 549: do we not care about the getTraceId part of the if condition anymore?

src/datasource.test.ts
```diff
@@ -1855,24 +1855,6 @@ describe('OpenSearchDatasource', function (this: any) {
         } as MetricAggregation,
       ],
     });
-
-    it('can handle multiple trace list queries', async () => {
```
iwysiu, line 1859: Do these tests have backend equivalents/are they just unnecessary now?

iwysiu (approved):
Looks good, a couple questions though

src/datasource.ts
idastambuk: We want all trace queries to go through the backend, whether they're trace list or single trace. 
Tbh it looks like even before this, all trace queries in Explore went through the backend because of the `metrics` field above (trace queries have a metric field "count", by default 😔 )

src/datasource.test.ts
idastambuk: It looks like there are some missing, added them!

pkg/opensearch/response_parser_test.go
```diff
@@ -2659,3 +2659,249 @@ func TestProcessTraceListResponse(t *testing.T) {
...
+
+	rp, err := newResponseParserForTest(targets, response, nil, client.ConfiguredFields{TimeField: "@timestamp"}, &backend.DataSourceInstanceSettings{UID: "123", Name: "DatasourceInstanceName"})
+	assert.Nil(t, err)
+
+	result, err := rp.parseResponse()
+	require.NoError(t, err)
+	require.Len(t, result.Responses, 2)
+
+	// 1st query
+	queryResTraceSpans1 := result.Responses["B"]
```
iwysiu, line 2765: Should one of these be `A`? Also totally optional, I know the original was very bare bones, but is there anything else we want to check in these to confirm parsing?

njvrzm (approved):

pkg/opensearch/response_parser_test.go
idastambuk: Done, added a check for operationName :)

PR #357, opened 2024-04-11
Rename ES Client to Client, cleanup 

idastambuk:
Just a small cleanup to eventually make the SG pull request smaller. Looks like a lot but it's mostly renaming 

- uses `client` or `os` instead of `es`, presumably `es` was a remnant of the elastic repo
- typos
- change type of Interval field to time.Duration and convert it only in getMinInterval


njvrzm (approved):
👍🏻 nice and straightforward

PR #358, opened 2024-04-12
Service Map: Add aggregations and query filters to use in service map queries

idastambuk:
Service Map feature: Add aggregations and query filters to use in service map queries

We're using some new aggregation types (`openSeachQuery.aggs`) and query filters (`openSearchQuery.query`) in requests for service map data to Open search. 

Query filters are what Open Search will filter the documents by, before applying the aggregations to the filtered set. 

**Query filters:** 
- MustNotFilters
- ShouldFilters (functions as OR for conditions inside it)
- Terms/Term filter (they must get marshaled differently depending on whether they have one or multiple values

**Aggregations:** 
- Script aggregation
- Terms aggregation (different than Terms filter) aggregating by term (field) `serviceName`

**Also:**
- tests
- renamed MustQueryBuilder to FilterList, as this struct wasn't Must-specific

**Note:**

The Bool filters that we need to add in order to get the right stats for the service map are quite nested and look a bit messy. This is because `filter`, `must _not`, `must`... filters always need to be nested under the `bool` filter, so there's a lot of repetition in the nesting. Once we start using them, we will need to have a complex query that looks like this:
 
Screenshot 2024-04-15 at 11 17 24

Part of the service map feature

Not using them in main, but they will be used in subsequent PRs for Service map

### DIFF
--- a/pkg/opensearch/client/models.go
+++ b/pkg/opensearch/client/models.go
@@ -264,11 +264,8 @@ type AggContainer struct {
 
 // MarshalJSON returns the JSON encoding of the aggregation container
 func (a *AggContainer) MarshalJSON() ([]byte, error) {
-	root := make(map[string]interface{})
-	if m, ok := a.Aggregation.(json.Marshaler); ok {
-		root[a.Type] = m
-	} else {
-		root[a.Type] = a.Aggregation
+	root := map[string]interface{}{
+		a.Type: a.Aggregation,
 	}
 
 	if len(a.Aggs) > 0 {
--- a/pkg/opensearch/client/models.go
+++ b/pkg/opensearch/client/models.go
@@ -100,10 +100,10 @@ type Query struct {
 
 // BoolQuery represents a bool query
 type BoolQuery struct {
-	Filters     []Filter
-	MustFilters []Filter
-	MustNotFilters []Filter `json:"must_not,omitempty"`
-	ShouldFilters  []Filter `json:"should,omitempty"`
+	Filters        []Filter
+	MustFilters    []Filter
+	MustNotFilters []Filter
+	ShouldFilters  []Filter
 }
 
 // MarshalJSON returns the JSON encoding of the boolean query.
@@ -336,12 +336,11 @@ type BucketScriptAggregation struct {
 	Script string            `json:"script"`
 }
 
-
 // TermsAggregation represents a terms aggregation
 type TermsAggregation struct {
 	Field       string                 `json:"field"`
 	Size        int                    `json:"size"`
-	Order       map[string]interface{} `json:"order,omitempty"`
+	Order       map[string]interface{} `json:"order"`
 	MinDocCount *int                   `json:"min_doc_count,omitempty"`
 	Missing     *string                `json:"missing,omitempty"`
 }
--- a/pkg/opensearch/client/models.go
+++ b/pkg/opensearch/client/models.go
@@ -325,7 +325,6 @@ func (f FilterAggregation) MarshalJSON() ([]byte, error) {
 			f.Key: f.Value,
 		},
 	}
-
 	return json.Marshal(root)
 }
 
@@ -372,7 +371,6 @@ func (a *MetricAggregation) MarshalJSON() ([]byte, error) {
 			root[k] = v
 		}
 	}
-
 	return json.Marshal(root)
 }
 
@@ -393,7 +391,6 @@ func (a *PipelineAggregation) MarshalJSON() ([]byte, error) {
 			root[k] = v
 		}
 	}
-
 	return json.Marshal(root)
 }
 
@@ -428,7 +425,6 @@ func (req *PPLRequest) MarshalJSON() ([]byte, error) {
 	root := map[string]interface{}{
 		"query": req.Query,
 	}
-
 	return json.Marshal(root)
 }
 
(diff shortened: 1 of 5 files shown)


pkg/opensearch/client/search_request_test.go
```diff
@@ -41,12 +41,68 @@ func TestSearchRequest(t *testing.T) {
...
+				assert.Equal(t, 200, sr.Size)
+			})
+			t.Run("Should have must not filter inside a should filter", func(t *testing.T) {
+				boolQuery, ok := sr.Query.Bool.Filters[0].(Query)
+				assert.True(t, ok)
+				mustNotFilters := boolQuery.Bool.ShouldFilters[0].(Query).Bool.Filters[0].(Query).Bool.MustNotFilters[0].(TermsFilter)
+				assert.Equal(t, "parentSpanId", mustNotFilters.Key)
+				assert.Equal(t, []string{""}, mustNotFilters.Values)
+			})
+			t.Run("Should have a terms filter inside a must not filter", func(t *testing.T) {
```
iwysiu, line 78: Is this test identical to the one right above it?

pkg/opensearch/client/search_request_test.go
```diff
@@ -41,12 +41,68 @@ func TestSearchRequest(t *testing.T) {
...
+				body, err := json.Marshal(sr)
+				assert.NoError(t, err)
+				json, err := simplejson.NewJson(body)
+				assert.NoError(t, err)
+
+				parentSpanId, err := json.GetPath("query", "bool", "filter", "bool", "should", "bool", "filter", "bool", "must_not", "term", "parentSpanId", "value").String()
+				assert.NoError(t, err)
+
+				assert.Equal(t, "", parentSpanId)
+
```
iwysiu, line 95: style nit: group function calls in the same code block as their asserts
```suggestion
				assert.NoError(t, err)
				assert.Equal(t, "", parentSpanId)
```

pkg/opensearch/client/search_request_test.go
```diff
@@ -179,28 +242,43 @@ func Test_Given_new_search_request_builder_for_es_OpenSearch_1_0_0(t *testing.T)
...
 			assert.NoError(t, err)
 			json, err := simplejson.NewJson(body)
 			assert.NoError(t, err)
 
-			assert.Len(t, json.Get("aggs").MustMap(), 2)
+			assert.Len(t, json.Get("aggs").MustMap(), 3)
 			assert.Equal(t, "@hostname", json.GetPath("aggs", "1", "terms", "field").MustString())
 			assert.Equal(t, "@timestamp", json.GetPath("aggs", "2", "date_histogram", "field").MustString())
+
+			// test for nested terms aggregations
```
iwysiu, line 278: Is this comment still necessary?

pkg/opensearch/client/search_request_test.go
```diff
@@ -179,28 +242,43 @@ func Test_Given_new_search_request_builder_for_es_OpenSearch_1_0_0(t *testing.T)
...
+
 		})
 	})
-
```
iwysiu, line 203: This line shouldn't be deleted. (as a style thing, it nice to have a new line between `t.Run` calls)

iwysiu (approved):
Nice! Some style nits

pkg/opensearch/client/models.go
```diff
@@ -156,6 +173,28 @@ func (f *QueryStringFilter) MarshalJSON() ([]byte, error) {
...
+}
+
+func (t TermsFilter) MarshalJSON() ([]byte, error) {
+	if len(t.Values) == 1 {
+		return json.Marshal(map[string]map[string]map[string]string{
+			"term": {
+				t.Key: {"value": t.Values[0]},
+			},
+		})
+	} else {
```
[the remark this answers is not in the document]
idastambuk: True, but I feel like adding the `else` explicitly is more readable 🤔

pkg/opensearch/client/search_request_test.go
idastambuk: Good catch, changed it to check for length

pkg/opensearch/client/search_request_test.go
idastambuk: Nope, added a test elsewhere

pkg/opensearch/client/models_test.go
```diff
@@ -0,0 +1,101 @@
...
+			assert.Nil(t, err)
+			assert.Equalf(t, tt.want, string(got), "MarshalJSON()")
+		})
+	}
+}
+
+func TestBoolQuery_MarshalJSON(t *testing.T) {
+	tests := []struct {
+		name    string
+		filters []Filter
```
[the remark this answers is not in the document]
idastambuk: Added some additional cases

pkg/opensearch/client/models.go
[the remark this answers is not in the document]
iwysiu: This is a short enough function that it doesn't really matter, but generally, the go style convention is to avoid having returns in elses, and `golint` actually will format your code that way (though we don't have `golint` turned on in our repos). Basically the idea for readability is that your functions are always returning at the end with the general behavior as left justified as possible, with exceptions branching off with an `if` (splitting off into sub functions if really necessary). 
(My old mentor's position was that you should never have an `else` in your go code ever, and while I'm not that extreme I do see his point that it ends up with a lot of different behavior paths that are easy to lose bugs in)
(Also If you want to read people talking about it on reddit here's a [link](https://www.reddit.com/r/golang/comments/e0p20u/golang_code_style_return_in_else_clause_what_do/), lol)

katebrenner:
fixes https://github.com/grafana/opensearch-datasource/issues/371

PR #362, opened 2024-04-16
Service Map: Implement service map index prefetch

idastambuk:
Split the feature into multiple PRs for ease of reviewing. The feature is not enabled by deafult but 'hidden' behind a provisional feature flag. 

In order to get stats for the service map (service info that's displayed inside a node: `avg latency`, `throughput`, `error rate`), we have to build an aggregation query on the `span` indices. The documents (spans) need to be filtered by operationNames and serviceNames, in order to get and aggregate *only* on spans that represent requests from one service to another (not internal processes). 
This reflects how the aggregations are build for the Service Map panel in Open Search. 

To get the operationNames and serviceNames, we have to first query the `service-map` indices and get operations and services from the `serviceName` and `destination.resource` fields of the index. 

**This PR implements:** 

1. detecting if the query requested from GF Open Search plugin has the serviceMap flag enabled
2. creating a separate query just for the `service-map` indices with the flag `serviceMapPrefetch`
3. adding flows that detect the `serviceMapPrefetch` flag and build the aggregations and the query
4. process the response from OpenSearch and store operation names and service names into Data Frames
5. putting those operations and services into the model of a service map Stats query and forwarding it along with the original query array (can contain other types of queries as well)
6. subsequent PRs will deal with building the stats query and processing it into final data frames that will be displayed in Grafana Node Graph visualization ([draft PR](https://github.com/grafana/opensearch-datasource/pull/363))

**Note:** This builds queries for otel + Open Search data prepper. Open Search can [store traces in raw Jaeger format](https://opensearch.org/docs/latest/observing-your-data/trace/index/#trace-analytics-with-jaeger-data), in which case the aggregations and responses should look slightly different. This is something we intend to implement in the future. 

**To test:**
1. enable `openSearchNodeGraph` feature toggle in Grafana config
2. run the Data Prepper/examples/jaeger-hotrod
3. generate some traces in the browser app
4. run the plugin in debugging mode
5. select query type: Traces and switch on the Service Map button in the query editor
6. you should be getting services and operations when you break on [this line](https://github.com/grafana/opensearch-datasource/blob/f6e1db87d81b77f8173703fa42ed8047c0c27c2e/pkg/opensearch/opensearch.go#L98)
7. nothing will show in the visualization yet, since we don't have stats processing yet

### DIFF
--- a/pkg/opensearch/client/search_request.go
+++ b/pkg/opensearch/client/search_request.go
@@ -422,6 +422,7 @@ const termsOrderTerm = "_term"
 func (b *aggBuilderImpl) Terms(key, field string, fn func(a *TermsAggregation, b AggBuilder)) AggBuilder {
 	innerAgg := &TermsAggregation{
 		Field: field,
+		Order: make(map[string]interface{}),
 	}
 	aggDef := newAggDef(key, &AggContainer{
 		Type:        "terms",
@@ -435,7 +436,6 @@ func (b *aggBuilderImpl) Terms(key, field string, fn func(a *TermsAggregation, b
 	}
 
 	if (b.version.Major() >= 6 || b.flavor == OpenSearch) && len(innerAgg.Order) > 0 {
-		innerAgg.Order = make(map[string]interface{})
 		if orderBy, exists := innerAgg.Order[termsOrderTerm]; exists {
 			innerAgg.Order["_key"] = orderBy
 			delete(innerAgg.Order, termsOrderTerm)
--- a/pkg/opensearch/client/search_request.go
+++ b/pkg/opensearch/client/search_request.go
@@ -168,6 +168,17 @@ func (b *SearchRequestBuilder) SetTraceListFilters(to, from int64, query string)
 	b.Size(10)
 }
 
+func (b *aggBuilderImpl) ServiceMap() AggBuilder {
+	b.Terms("service_name", "serviceName", func(a *TermsAggregation, b AggBuilder) {
+		b.Terms("destination_domain", "destination.domain", func(a *TermsAggregation, b AggBuilder) { b.Terms("destination_resource", "destination.resource", nil) })
+
+		b.Terms("target_domain", "target.domain", func(a *TermsAggregation, b AggBuilder) {
+			b.Terms("target_resource", "target.resource", nil)
+		})
+	})
+	return b
+}
+
 // Search initiates and returns a new search request builder
 func (m *MultiSearchRequestBuilder) Search(interval tsdb.Interval) *SearchRequestBuilder {
 	b := NewSearchRequestBuilder(m.flavor, m.version, interval)
@@ -328,6 +339,7 @@ type AggBuilder interface {
 	Terms(key, field string, fn func(a *TermsAggregation, b AggBuilder)) AggBuilder
 	Filters(key string, fn func(a *FiltersAggregation, b AggBuilder)) AggBuilder
 	TraceList() AggBuilder
+	ServiceMap() AggBuilder
 	GeoHashGrid(key, field string, fn func(a *GeoHashGridAggregation, b AggBuilder)) AggBuilder
 	Metric(key, metricType, field string, fn func(a *MetricAggregation)) AggBuilder
 	Pipeline(key, pipelineType string, bucketPath interface{}, fn func(a *PipelineAggregation)) AggBuilder
@@ -422,7 +434,6 @@ const termsOrderTerm = "_term"
 func (b *aggBuilderImpl) Terms(key, field string, fn func(a *TermsAggregation, b AggBuilder)) AggBuilder {
 	innerAgg := &TermsAggregation{
 		Field: field,
-		Order: make(map[string]interface{}),
 	}
 	aggDef := newAggDef(key, &AggContainer{
 		Type:        "terms",
--- a/pkg/opensearch/client/search_request.go
+++ b/pkg/opensearch/client/search_request.go
@@ -434,6 +434,7 @@ const termsOrderTerm = "_term"
 func (b *aggBuilderImpl) Terms(key, field string, fn func(a *TermsAggregation, b AggBuilder)) AggBuilder {
 	innerAgg := &TermsAggregation{
 		Field: field,
+		Size: 500,
 	}
 	aggDef := newAggDef(key, &AggContainer{
 		Type:        "terms",
--- a/pkg/opensearch/client/search_request.go
+++ b/pkg/opensearch/client/search_request.go
@@ -170,9 +170,18 @@ func (b *SearchRequestBuilder) SetTraceListFilters(to, from int64, query string)
 
 func (b *aggBuilderImpl) ServiceMap() AggBuilder {
 	b.Terms("service_name", "serviceName", func(a *TermsAggregation, b AggBuilder) {
-		b.Terms("destination_domain", "destination.domain", func(a *TermsAggregation, b AggBuilder) { b.Terms("destination_resource", "destination.resource", nil) })
+		a.Size = 500
+		b.Terms("destination_domain", "destination.domain", func(a *TermsAggregation, b AggBuilder) {
+			a.Size = 500
+			b.Terms("destination_resource", "destination.resource", func(a *TermsAggregation, b AggBuilder) {
+				a.Size = 500
+			})
+		})
 		b.Terms("target_domain", "target.domain", func(a *TermsAggregation, b AggBuilder) {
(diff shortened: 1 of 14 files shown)


pkg/opensearch/lucene_handler.go
```diff
@@ -57,26 +59,31 @@ func (h *luceneHandler) processQuery(q *Query) error {
...
 	if q.luceneQueryType == luceneQueryTypeTraces {
 		traceId := getTraceId(q.RawQuery)
-		if traceId != "" {
-			b.Size(1000)
-			b.SetTraceSpansFilters(toMs, fromMs, traceId)
-		} else {
-			b.SetTraceListFilters(toMs, fromMs, q.RawQuery)
+		switch q.serviceMapInfo.Type {
+		case Prefetch:
+			b.Size(0)
```
sarahzinger, line 69: kinda silly question on my part, why is this zero? Like we want to get some amount of results no?

pkg/opensearch/lucene_handler.go
```diff
@@ -284,6 +290,32 @@ func (h *luceneHandler) executeQueries(ctx context.Context) (*backend.QueryDataR
...
+	buckets := smResult.Aggregations["service_name"].(map[string]interface{})["buckets"].([]interface{})
+	for _, bucket := range buckets {
+		service := bucket.(map[string]interface{})
+
+		services = append(services, service["key"].(string))
+		targets := service["target_domain"].(map[string]interface{})["buckets"].([]interface{})
+		for _, targetDomain := range targets {
+			targetResources := targetDomain.(map[string]interface{})["target_resource"].(map[string]interface{})["buckets"].([]interface{})
+			for _, targetResource := range targetResources {
+				operationMap[targetResource.(map[string]interface{})["key"].(string)] = true
```
sarahzinger, line 308: I don't know golang well enough to say this with full confidence but it feels like there must be a way somehow to do this that is not constantly casting stuff, or maybe this is just how golang is lol? Then again I feel like that could be fixed in a subsequent pr probs since this is all just in one little func.

pkg/opensearch/client/search_request.go
```diff
@@ -422,7 +433,7 @@ const termsOrderTerm = "_term"
 func (b *aggBuilderImpl) Terms(key, field string, fn func(a *TermsAggregation, b AggBuilder)) AggBuilder {
 	innerAgg := &TermsAggregation{
 		Field: field,
-		Order: make(map[string]interface{}),
+		Size:  500,
```
sarahzinger, line 436: should the size be configurable? Maybe we can make a ticket for that later... also is there any risk this could like cause a breaking change for existing Term searches? Like If I wanted to group by a term and get the top 10 results, this wouldn't somehow cause me to get 500 right?

pkg/opensearch/opensearch.go
```diff
@@ -64,11 +65,53 @@ func (ds *OpenSearchDatasource) QueryData(ctx context.Context, req *backend.Quer
...
+		serviceMapRequested := model.Get("serviceMap").MustBool(false)
+		if queryType == Lucene && luceneQueryType == "Traces" && serviceMapRequested {
+			serviceMapQueryIndex = i
+			serviceMapQuery = createServiceMapPrefetchQuery(query)
+			break
+		}
+	}
+	if serviceMapQuery.JSON != nil {
+		query := newQueryRequest(osClient, []backend.DataQuery{serviceMapQuery}, req.PluginContext.DataSourceInstanceSettings, intervalCalculator)
+		response, _ := wrapError(query.execute(ctx))
```
sarahzinger, line 98: if the prefetch errors, do we return the response as an error? It looks like wrapError doesn't always send back the response? I am probably misunderstanding but it looks like we swallow the error here?

pkg/opensearch/opensearch.go
```diff
@@ -64,11 +65,53 @@ func (ds *OpenSearchDatasource) QueryData(ctx context.Context, req *backend.Quer
...
+		services, operations := extractParametersFromServiceMapFrames(response)
+
+		// encode the services and operations back to the JSON of the query to be used in the stats request
+		model, _ := simplejson.NewJson(req.Queries[serviceMapQueryIndex].JSON)
```
sarahzinger, line 102: should we use `_` here? Seems like we could conceivably have an error here and deal with it?

pkg/opensearch/opensearch_test.go
```diff
@@ -30,3 +35,93 @@ func Test_wrapError(t *testing.T) {
 		assert.Equal(t, "OpenSearch data source error: some error", err.Error())
 	})
 }
+
+func TestServiceMapPreFetch(t *testing.T) {
+	targets1 := map[string]interface{}{
```
sarahzinger, line 40: this is definitely a "sarah doesn't understand go" question, but is there a way to rewrite this so we're looking at a json object or struct something so it's easier to read? Or is that like "not done" really?

sarahzinger (approved):
Awesome work! I have some questions mostly for my own understanding although it's possible some might be helpful? haha. Feel free to re-request a review if you need, but also trust y'all to do what seems best here.

pkg/opensearch/client/search_request.go
```diff
@@ -168,6 +168,16 @@ func (b *SearchRequestBuilder) SetTraceListFilters(to, from int64, query string)
 	b.Size(10)
 }
 
+func (b *aggBuilderImpl) ServiceMap() AggBuilder {
```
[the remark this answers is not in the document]
idastambuk: We could hardcode the object, yes. We do something similar with Trace List. 
We decided to work with the functions that we already have when it's simple enough to use them (for no reason i particular, but correct me if I'm wrong @njvrzm). We're keeping query building in mind for a refactor at some point to make it more readable though

pkg/opensearch/lucene_handler.go
idastambuk: The size parameter is used to determine how many `hits` are returned in the response. Meaning, documents that match the whatever filters, range etc. the query sends. Since we only need aggregations (`buckets` field) in this response, we don't need or use the `hits` field.

pkg/opensearch/client/search_request.go
idastambuk: You're totally right, I moved the size into the fn() where Terms aggregation is added

pkg/opensearch/opensearch.go
idastambuk: I don't think we expect an error here, since it's a non-modified request coming from FE, but it can't hurt to add it!

pkg/opensearch/client/search_request.go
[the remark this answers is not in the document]
njvrzm: That's what I recall as well, and yeah, there's a lot of room here for improvement on the query building code.

pkg/opensearch/lucene_handler.go
njvrzm: I have a test branch somewhere that provides helpers to do this in a nicer way, would love to get that in soon!

pkg/opensearch/opensearch.go
njvrzm: Yeah, no, you're not misunderstanding 😅 I think we just missed this. It's handled now.

pkg/opensearch/opensearch.go
njvrzm: In this case it is impossible for there to be an error - we've already called `simplejson.NewJson` on the `JSON` for every query, including this one, on line 86. I like leaving the `_` here to indicate that, but I can add a comment as well.

njvrzm: (Well - impossible barring concurrency shenanigans, which I don't _think_ can happen here)

pkg/opensearch/opensearch_test.go
njvrzm: We can totally write these in JSON and then decode them - that will be more readable indeed.

idastambuk: Done!

pkg/opensearch/opensearch.go
idastambuk: We kinda still swallow the error tbh. As @sarahzinger mentioned, if we don't return the error in the Response field, it won't be shown to the end user. We started by writing the wrapError function to make it possible to return custom errors, but from tinkering with it a little bit this morning, I don't see a simple and pretty way to do it for the prefetch. Since this is the same problem we have with all the other query types (they all show "downstream plugin error") I would suggest we add this as an improvement later, if you agree?

katebrenner:
closes https://github.com/grafana/opensearch-datasource/issues/370

PR #367, opened 2024-05-01
Backend Migration: migrate ppl timeseries to backend

### LINKED ISSUE #296: Backend: Migrate PPL Time Series to the backend
Prerequisite: https://github.com/grafana/opensearch-datasource/issues/226

Some general guidelines are:

* Create visualization to compare between frontend and backend flows, preferably in a shared Cloud dashboard e.g. clouddatasources.grafana.net 
* Unit tests in the frontend should be copied or updated in the backend to assert the same behavior and add unit tests when there are none.
    * Check Elasticsearch repo for related tests – helpful when there are no OpenSearch tests in the frontend or backend
* We should match OpenSearch's backend behavior to the current frontend behavior, but be aware it may not always make sense. Elasticsearch's current behavior is also another resource for how OpenSearch should behave.
    * Let's be pragmatic about this – our foremost goal is the migration of OpenSearch's frontend behavior. We can aim for some Elasticsearch feature parity, but only _within reason_.
* Often there is a PR in the Elasticsearch backend data source migration [https://github.com/grafana/grafana/issues/54011](https://github.com/grafana/grafana/issues/54011) which is 75% of what OpenSearch will need
* Migrate or implement: i.e. make any changes to code in backend
* Use the dashboard created above to compare visualization results (data frame comparison)
    * Can also inspect data frame on a more detailed basis

iwysiu:
The code for the backend ppl timeseries actually already matched up pretty well with the frontend (except the field name) and actually had more tests, so this is more of a making-it-a-bit-nicer pr.

Fixes #296 

### DIFF
--- a/pkg/opensearch/ppl_response_parser.go
+++ b/pkg/opensearch/ppl_response_parser.go
@@ -17,7 +17,7 @@ type pplResponseParser struct {
 	Response *client.PPLResponse
 }
 
-func newPPLResponseParser(response *client.PPLResponse, query *Query) *pplResponseParser {
+func newPPLResponseParser(response *client.PPLResponse) *pplResponseParser {
 	return &pplResponseParser{
 		Response: response,
 	}
@@ -142,10 +142,12 @@ func (rp *pplResponseParser) parseTimeSeries(queryRes *backend.DataResponse) (*b
 		return nil, err
 	}
 
-	newFrame := data.NewFrame(rp.getSeriesName(t.valueIndex),
+	valueName := rp.getSeriesName(t.valueIndex)
+	newFrame := data.NewFrame(valueName,
 		data.NewFieldFromFieldType(data.FieldTypeNullableTime, len(rp.Response.Datarows)),
 		data.NewFieldFromFieldType(data.FieldTypeNullableFloat64, len(rp.Response.Datarows)),
 	)
+	newFrame.Fields[1].Name = valueName
 
 	for i, datarow := range rp.Response.Datarows {
 		err := rp.addDatarow(newFrame, i, datarow, t)
@@ -155,7 +157,6 @@ func (rp *pplResponseParser) parseTimeSeries(queryRes *backend.DataResponse) (*b
 	}
 
 	queryRes.Frames = append(queryRes.Frames, newFrame)
-
 	return queryRes, nil
 }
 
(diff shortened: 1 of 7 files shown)


pkg/opensearch/ppl_response_parser.go
```diff
@@ -17,7 +17,7 @@ type pplResponseParser struct {
 	Response *client.PPLResponse
 }
 
-func newPPLResponseParser(response *client.PPLResponse, query *Query) *pplResponseParser {
+func newPPLResponseParser(response *client.PPLResponse) *pplResponseParser {
```
iwysiu, line 20: I can change this back, but it was annoying me and my formatter. (this is a private function so its fine to change the funcition signature)

pkg/opensearch/ppl_response_parser.go
```diff
@@ -224,7 +226,12 @@ func getTimeSeriesResponseMeta(schema []client.FieldSchema) (responseMeta, error
 	if !found {
 		return responseMeta{}, errors.New("a valid time field type was not found in response")
 	}
-	return responseMeta{valueIndex: 1 - timeIndex, timeFieldIndex: timeIndex, timeFieldFormat: format}, nil
+
+	var valueIndex int
+	if timeIndex == 0 {
```
iwysiu, line 231: Because of how the rest of the function logic works this actually gets the same value as the old code, but its a little easier to read

pkg/opensearch/ppl_response_parser_test.go
```diff
@@ -212,7 +214,7 @@ func TestPPLResponseParser(t *testing.T) {
 				rp, err := newPPLResponseParserForTest(targets, response)
 				assert.NoError(t, err)
 				_, err = rp.parseResponse(client.ConfiguredFields{}, "")
-				assert.Error(t, err)
+				assert.ErrorContains(t, err, "response should have 2 fields but found 3")
```
iwysiu, line 217: I thought it would be nice to check for more specific errors

pkg/opensearch/ppl_response_parser_test.go
```diff
@@ -295,18 +297,18 @@ func TestPPLResponseParser(t *testing.T) {
 				response := `{
 							"schema": [
 								{ "name": "timeName", "type": "timestamp" },
-								{ "name": "testMetric", "type": "string" }
+								{ "name": "testMetric", "type": "integer" }
```
iwysiu, line 300: With an invalid value type, this and the test below weren't actually getting to the part of the code that tests the timestamp

kevinwcyu (approved):

PR #369, opened 2024-05-01
Fix flaky tests

### LINKED ISSUE #368: Fix flaky node graph tests
**What happened**:
Two of the node graph tests added in [test 1](https://github.com/grafana/opensearch-datasource/blob/8b2f845fe03cd6d9f3c20e42ebd0227552a2a4e4/pkg/opensearch/response_parser_test.go#L2667) and [test 2](https://github.com/grafana/opensearch-datasource/blob/8b2f845fe03cd6d9f3c20e42ebd0227552a2a4e4/pkg/opensearch/opensearch_test.go#L39) are flaky due to the map used by backend.Response.

**Environment**:
- Grafana version:
- OpenSearch version:
- Plugin version:

iwysiu:
Due to backend.Responses being a map, we end up getting the services/operations in an inconsistent order when we have multiple responses. This makes some of the tests flaky ([test 1](https://github.com/grafana/opensearch-datasource/blob/8b2f845fe03cd6d9f3c20e42ebd0227552a2a4e4/pkg/opensearch/response_parser_test.go#L2667) and [test 2](https://github.com/grafana/opensearch-datasource/blob/8b2f845fe03cd6d9f3c20e42ebd0227552a2a4e4/pkg/opensearch/opensearch_test.go#L39)). One fix for that is just sorting the operations and services (done in this pr), though if we think that's too expensive (I don't think we should have that many of them?) we could also try to rework the tests, though considering that one of them involves checking a big json string that could either be a bit complex or just involve not using 2 responses (though that reduces the test coverage).

The other change was to avoid a seg fault in `TestProcessTraceListAndTraceSpansResponse`, which was happening because the helper that generated the test queries used a map and was unordered. Changed to an array.

Fixes #368 

### DIFF
--- a/pkg/opensearch/lucene_handler.go
+++ b/pkg/opensearch/lucene_handler.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"regexp"
-	"slices"
 	"strconv"
 	"strings"
 	"time"
@@ -315,7 +314,6 @@ func getParametersFromServiceMapResult(smResult *client.SearchResponse) ([]strin
 	for op := range operationMap {
 		operations = append(operations, op)
 	}
-	slices.Sort(operations)
 	return services, operations
 }
 
--- a/pkg/opensearch/lucene_handler.go
+++ b/pkg/opensearch/lucene_handler.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
@@ -309,10 +310,12 @@ func getParametersFromServiceMapResult(smResult *client.SearchResponse) ([]strin
 			}
 		}
 	}
+
 	operations := make([]string, 0, len(operationMap))
 	for op := range operationMap {
 		operations = append(operations, op)
 	}
+	slices.Sort(operations)
 	return services, operations
 }
 
--- a/pkg/opensearch/response_parser_test.go
+++ b/pkg/opensearch/response_parser_test.go
@@ -17,15 +17,20 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func createDataQueriesForTests(tsdbQueries map[string]string) []backend.DataQuery {
+type tsdbQuery struct {
+	refId string
+	body  string
+}
+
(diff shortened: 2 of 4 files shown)


pkg/opensearch/response_parser_test.go
```diff
@@ -17,15 +17,20 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func createDataQueriesForTests(tsdbQueries map[string]string) []backend.DataQuery {
+type tsdbQuery struct {
+	refId string
+	body  string
+}
+
+func createDataQueriesForTests(tsdbQueries []tsdbQuery) []backend.DataQuery {
```
iwysiu, line 25: Alternatively I could create a second helper that takes an array, but the map sort of implies that it should take multiple queries.

pkg/opensearch/response_parser_test.go
```diff
@@ -2831,53 +2879,49 @@ func TestProcessTraceListAndTraceSpansResponse(t *testing.T) {
...
-								},
-								"traceGroup": "HTTP GET /dispatch",
-								"serviceName": "frontend",
-								"parentSpanId": "",
-								"spanId": "47ed3a25a7dba0cd",
-								"traceState": "",
-								"name": "HTTP GET /dispatch",
-								"startTime": "2023-10-18T07:58:37.818589Z",
-								"links": [],
+			}
```
iwysiu, line 2882: I just changed the indents on this to be easier to read.

pkg/opensearch/lucene_handler.go
```diff
@@ -309,10 +310,12 @@ func getParametersFromServiceMapResult(smResult *client.SearchResponse) ([]strin
 			}
 		}
 	}
+
 	operations := make([]string, 0, len(operationMap))
 	for op := range operationMap {
 		operations = append(operations, op)
 	}
+	slices.Sort(operations)
```
sarahzinger, line 318: I feel like if sorting the operations benefitted users in someway it might make sense to do this, but if I understand correctly this is always going to be rendered as a graph where sorting won't really matter. I would vote we try sorting in the test if that's possible?

Issue #374, opened 2024-05-06
Backend Migration: Remove PPL Time Series Explore-only restriction (condition is in frontend)


kevinwcyu:

kevinwcyu:
Will be done in https://github.com/grafana/opensearch-datasource/issues/407

PR #375, opened 2024-05-06
Run all Lucene queries and PPL logs and table queries on the backend

### LINKED ISSUE #275: Backend Migration: Remove Lucene Logs Explore-only restriction (condition is in frontend)

### LINKED ISSUE #290: Backend Migration: Remove Lucene Metrics Explore-only restriction (condition is in frontend)

### LINKED ISSUE #372: Backend Migration: Remove PPL Logs Explore-only restriction (condition is in frontend)

### LINKED ISSUE #373: Backend Migration: Remove PPL Table Explore-only restriction (condition is in frontend)

kevinwcyu:
All Lucene queries and PPL logs and table queries now run through the backend.

Fixes #275
Fixes #290
Fixes #372
Fixes #373

I left the condition to run PPL time series queries on the backend only in Explore because that was only merged recently in #367. I created a ticket to follow up on removing the condition for PPL time series here https://github.com/grafana/opensearch-datasource/issues/374.


src/datasource.ts
```diff
@@ -535,14 +535,8 @@ export class OpenSearchDatasource extends DataSourceWithBackend<OpenSearchQuery,
     if (
       request.targets.every(
         (target) =>
-          target.metrics?.every(
-            (metric) =>
-              metric.type === 'raw_data' ||
```
sarahzinger, line 540: I think we are losing these types `raw_data` and `raw_document` or am I not reading this right? I wonder if it would be easier at this point to say what doesn't go through the backend flow rather than what does?

sarahzinger (approved):
Thanks for working on this! I think you've got a failing test and I have a (probably stupid question) but once addressed and once the tests pass I'm cool for this to get merged

src/datasource.ts
kevinwcyu: The `raw_data` and `raw_document` are Lucene queries that we never restricted to execute on the backend in Explore only.

I've updated with your suggestion to check what shouldn't go through the backend flow, which should be
1. PPL query
2. formatted as a time series
3. run from the dashboard

src/datasource.test.ts
```diff
@@ -142,260 +137,6 @@ describe('OpenSearchDatasource', function (this: any) {
...
-      expect(result.data[0].name).toEqual('resolvedVariable');
-    });
-
-    it('should json escape lucene query', () => {
-      const body = JSON.parse(parts[1]);
-      expect(body.query.bool.filter[1].query_string.query).toBe('escape\\:test');
-    });
-  });
-
-  describe('When using sigV4 and POST request', () => {
```
kevinwcyu, line 219: Moved these tests inside of the `PPL Queries` test block since only PPL time series format queries are still run on the frontend in the dashboard.

kevinwcyu (commented):
@sarahzinger, I've fixed the failing tests by removing them. They are for the code that we no longer execute on the frontend now that we're sending all of the queries to the backend. The removed tests were testing how we query and parse the response directly from the OpenSearch API, but that is now handled in the backend.

LoLZeS666:
Hi, I think updating the plugin to v2.15.0 also broke some queries for me. All queries where I used 'group by' on multiple terms, only the last group by would work. Reverting to v2.14.7 fixed the issue. Not sure if it related to this change or not.
Grafana v10.4.2
Opensearch v2.13.0
Plugin v2.15.0

idastambuk:
Hi @LoLZeS666 thanks for the info! We've reverted the migration in patch v12.15.1 until we fix the issue.

PR #378, opened 2024-05-10
Service Map: Add tests for stats request

idastambuk:

Fixes #


pkg/opensearch/snapshot_tests/lucene_service_map_prefetch_test.go
```diff
@@ -1,49 +0,0 @@
-package snapshot_tests
```
idastambuk, line 1: Moved this to lucene_service_map_test.go

pkg/opensearch/lucene_handler.go
```diff
@@ -318,6 +319,10 @@ func getParametersFromServiceMapResult(smResult *client.SearchResponse) ([]strin
 	for op := range operationMap {
 		operations = append(operations, op)
 	}
+	// ensure consistent order for the snapshot tests in lucene_service_map_test.go
```
idastambuk, line 322: Don't love this solution just to get the snapshot tests to pass. But it feels like it would be a lot of extra code to sort the response in the test (extract the parts of the multiline json that send the services and operations list + sort the order + marshall it back to json)

njvrzm: I think this is well in the realm of "not pretty but fine".

njvrzm (approved):
Looks good to me :)

Issue #389, opened 2024-05-15
Unable to parse interval


wilfriedroset:
**What happened**:

All panels relying on opensearch plugins and using a standard interval variable encounter an error.

https://github.com/grafana/opensearch-datasource/assets/12611310/13bdb390-d3c0-491e-aead-94eb6faae9da

```
Unable to parse interval [$interval]
```

**What you expected to happen**:

The panels load without any error and the variable used for the interval is replace before sending the request to the backend.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:
The bug seems to have been introduced in either #375 or #367

**Environment**:
- Grafana version: 10.1.9
- OpenSearch version: 1.3.14
- Plugin version: 1.5.0

iwysiu:
Hi @wilfriedroset ! Thanks for reporting this issue, I've reproduced it and I've moved it to our next column to be looked at soon.

idastambuk:
Hi @wilfriedroset and @tstm, we have reverted the migration with the version 2.15.1 so you can update to that version to fix your dashboards until we're able to solve the problem

PR #399, opened 2024-05-27
Refactor ad hoc variable processing 

idastambuk:
Previously, we were interpolating variables in the `query()` method manually. This adds several methods to the datasource that are called by code Grafana whenever variable and ad hoc interpolation is necessary: 

- `applyTemplateVariables` - this is a method that is called by Grafana during the super.query() call, which means we don't have to do this manually in our datasource. The method implements template variable interpolation as well as ad hoc queries following the approach in Elastic search from [this PR](https://github.com/grafana/grafana/pull/65274)
- `interpolateVariablesInQueries` called from multiple places in core grafana, didn't look too much into it but most datasources implement it. Elastic [here](https://github.com/grafana/grafana/blob/main/public/app/plugins/datasource/elasticsearch/datasource.ts#L431)
- copied processing for ad hoc filters for lucene queries added in Elastic [in this PR](https://github.com/grafana/grafana/pull/71954) 
- added processing for PPL ad hoc filters. This is mostly copied from what we had, with the difference of adding removeFilterFromPPLQuery, PPLQueryHasFilter, toggleQueryFilterForPPL etc. that are now used for both adding ad hoc filters in dashboards and in Explore
- also fixes https://github.com/grafana/opensearch-datasource/issues/389 by [interpolating the stringified query](https://github.com/grafana/opensearch-datasource/blob/1dc25998bdf13e82ad51b2ff05f3ee983e827dec/src/datasource.ts#L412)

Fixes https://github.com/grafana/opensearch-datasource/issues/379, https://github.com/grafana/opensearch-datasource/issues/389

- this was mostly tested with backend flow, but the frontend flow seems to be improved as well, mainly when adding ad hoc queries in Explore

**How to test**
1. In Dashboard run Lucene "Raw data" query with table panel and click on magnifying glasses in the table cells. 
- `+` adds a filter
- `+` again removes a filter
- `-` filters out a field.
2. Make sure that executed query (using query inspector) has filters as `query.query`. The query won't have filters in the UI, just the query inspector
3. Switch to Explore and do the filtering in the same way. In Explore, the query should be shown in the UI Query field
4. Repeat 1-3 for PPL queries


src/datasource.ts
```diff
@@ -568,6 +551,20 @@ export class OpenSearchDatasource extends DataSourceWithBackend<OpenSearchQuery,
     }
 
     // Frontend flow
+    const targetsWithInterpolatedVariables = request.targets.map((target) => {
```
idastambuk, line 554: Just added interpolating template variables because adhoc filters are later added for Lucene and PPL in the frontend flow, e.g.
`this.queryBuilder.buildPPLQuery(target, adhocFilters, queryString)`

iwysiu (commented):
When I hit `+` twice for a value in the dashboard panel it didn't remove the filter, but in Explore it did. Is that an expected difference (because of the migration) or is that a bug?

idastambuk:
> When I hit `+` twice for a value in the dashboard panel it didn't remove the filter, but in Explore it did. Is that an expected difference (because of the migration) or is that a bug?

This is expected, should have mentioned it. Grafana in those cases still passes the same filters without removing existing ones. I assume this is because you can remove the filters from the variable list above the panels.

src/modifyQuery.ts
```diff
@@ -0,0 +1,362 @@
...
+        : addAdhocFilterToPPLQuery(expression, adHocFilter);
+      break;
+    }
+    case 'FILTER_OUT': {
+      const adHocFilter: AdHocVariableFilter = {
+        key: filter.options.key,
+        value: filter.options.value,
+        operator: '!=',
+      };
+      // If the opposite filter is present, remove it before adding the new one.
```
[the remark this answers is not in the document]
idastambuk: As far as I understand, that's the case for ad hoc filters, yes. Filter out is basically "only show != value"

src/datasource.test.ts
```diff
@@ -2111,6 +1993,134 @@ describe('OpenSearchDatasource', function (this: any) {
...
+        metrics: [{ type: 'count', id: '1' }],
+        query: '$var',
+      };
+
+      const interpolatedQueries = ctx.ds.interpolateVariablesInQueries([query], {}, adHocFilters);
+      expect(interpolatedQueries[0].query).toBe(
+        "resolvedVariable | where `bar` = 'baz' and `job` != 'grafana' and `bytes` > 50 and `count` < 100 and `timestamp` = timestamp('2020-11-22 16:40:43.000000')"
+      );
+    });
+  });
```
iwysiu, line 2034: nit: newlines before and after this describe block

src/datasource.test.ts
```diff
@@ -2031,34 +1899,48 @@ describe('OpenSearchDatasource', function (this: any) {
...
       });
 
-      it('should escape characters in filter keys', () => {
+      it('PPL queries should correctly add 1 ad hoc filter when query is empty', () => {
+        // an empty string query is transformed to '*' but this can be refactored to have the same behavior as Elasticsearch
+        const query = ctx.ds.addAdHocFilters({ refId: 'A', query: '', queryType: QueryType.PPL }, adHocFilters);
+        expect(query).toBe("`test` = 'test1'");
+      });
+    });
+    describe('Escaping characters in adhoc filter', () => {
```
iwysiu, line 1942: nit: newline before and after this describe block

iwysiu (approved):
Just a couple more new lines

kevinwcyu (approved):

Issue #388, opened 2024-05-14
Opensearch Datasource Dashboard Facing Issues: An error occurred within the plugin


SricharanSundar:
**What happened**:
   The Opensearch datasource dashboard is not working properly. Its giving an error occurred within the plugin.
   Im getting the following output in query response.
   
   traceId:undefined
  request:Object
  url:"api/ds/query?ds_type=grafana-opensearch-datasource&requestId=Q234"
  method:"POST"
  data:Object
  queries:Array[1]
  from:"1715583459199"
  to:"1715669859200"
  hideFromInspector:false
  response:Object
  message:"An error occurred within the plugin"
  messageId:"plugin.downstreamError"
  statusCode:500
  traceID:""
   
**What you expected to happen**:
   The dashboard need to work properly always. Few days before the dashboard work properly without any error but now it is causing issues. Now a days grafana releases many version but its creating some issues in existing dashboard.
   
**How to reproduce it (as minimally and precisely as possible)**:
   connect the opensearch data source with opensearch url endpoints and use sigv4 auth method.
   Then create on log based dashboard using bar chart visualization. 
   Once the dashboard is saved, then sign out  grafana and sign in again youre able to see this issue.
   
**Anything else we need to know?**:
   Fix this issue as soon as possible in grafana.
   
**Environment**:
- Grafana version: 10.4.1
- OpenSearch version: 2.9
- Plugin version: Latest version.

iwysiu:
Hi @SricharanSundar ! I'm currently looking into this, I have a couple questions to help my investigation:
- What's the last version that it worked properly on?
- If you go to the the configured datasource and click test, does it succeed or fail? And if it fails, what's the error?
- If you have access to the backend, can you get the error from the logs?

SricharanSundar:
Hi @iwysiu,

What's the last version that it worked properly on?
Answer: The dashboard is worked properly on the same version of Grafana for few days before. But suddenly its giving issues in the panel.
Grafana version: 10.4.1
Opensearch: 2.9

If you go to the the configured datasource and click test, does it succeed or fail? And if it fails, what's the error?
The configuration is success and the dashboard is working for few peoples in a team. For few peoples we are facing this issue(An error occurred within the plugin). 
Most of the team members cant able to view the dashboard due to this issue.

If you have access to the backend, can you get the error from the logs?
status:500
statusText:"Internal Server Error"
data:Object
message:"An error occurred within the plugin"
messageId:"plugin.downstreamError"
[...]

PR #404, opened 2024-06-05
Bugfix: Pass docvalue_fields for elasticsearch

idastambuk:
We had a bug reported in https://github.com/grafana/opensearch-datasource/issues/388#issuecomment-2125144165 where logs queries to ElasticSearch v7.8.1 with OpenSearch plugin were returning No Data. The reason for this was that we're sending "fields" [here](https://github.com/grafana/opensearch-datasource/blob/c8067fcf6c7ed0e30cad4854869286c8ca7f61be/pkg/opensearch/client/search_request.go#L110) to OpenSearch with the timestamp format we desire. 
ElasticSearch api, however, expects `docvalue_fields` instead, so it returns no data. Older versions of Elastic expect `fields` as well. 

There was also some refactoring, since without it, the time field was in some cases added twice in `fields` or `docvalue_fields` (e.g. `docvalue_fields: ["@timestamp", {"field": "@timestamp", "format": "date_time"}]`
- Moved adding the `timeField` to `AddTimeFieldWithStandardizedFormat`, where the field, along with its formatting will be added according to the flavor and version. Still not sure if `fielddata_fields` is completely necessary, but since it's only for Elastic < 5 and logs queries, it might not be relevant in time if we decide to deprecate. The same goes for the new version of `AddDocValueFields` function
- removed `script_fields` field since it doesn't seem to be used anywhere

Fixes https://github.com/grafana/opensearch-datasource/issues/405

To test with Elastic 7.8.1: 
- run make devenv sources=elasticstack elastic_version=7.8. and make run in main Grafana
- open http://localhost:5601/app/kibana#/home/tutorial_directory/logging and add some sample data (Sample web logs)
- add a new OpenS datasource with this config

- make a logs query/raw document query
- there should be data shown

### DIFF
--- a/pkg/opensearch/client/search_request.go
+++ b/pkg/opensearch/client/search_request.go
@@ -93,40 +93,36 @@ func (b *SearchRequestBuilder) Sort(order, field, unmappedType string) *SearchRe
 	return b
 }
 
-// AddDocValueField adds a doc value field to the search request
-func (b *SearchRequestBuilder) AddDocValueField(field string) *SearchRequestBuilder {
-	b.customProps["script_fields"] = make(map[string]interface{})
-
+func (b *SearchRequestBuilder) AppendDocValueFields(field string) {
 	if b.version.Major() < 5 && b.flavor == Elasticsearch {
-		b.customProps["fielddata_fields"] = []any{field}
-		b.customProps["fields"] = []any{"*", "_source"}
-	} else {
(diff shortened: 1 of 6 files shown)


kevinwcyu (approved):
LGTM. Do you know why it worked in the frontend, but not the backend path? I didn't see any of these fields in the frontend code.

kevinwcyu (approved):

PR #409, opened 2024-06-14
Backend migration: Send all queries to backend if openSearchBackendFlow toggle is enabled

idastambuk:

Gradually roll out backend migration with the feature toggle. 
Decided to keep the Explore queries also on backend flow, since that has been enabled for a long while and I didn't want to change functionality. 
Also refactored the tests a bit to reduce repetition. 

Fixes #


sarahzinger (approved):
🎉

PR #418, opened 2024-07-09
Bring back the selector for opensearch version

### LINKED ISSUE #206: Automatic Version detection sometimes gets Elasticsearch 7.10.2 for Opensearch
Some users are reporting that when they use the `Save and Set Version` button in the config page, the version that is found is Elasticsearch 7.10.2 despite being on Opensearch.

After investigating, I found that when this was happening the Opensearch instance was responding to the version request with the response for an Elasticsearch 7.10.2. I haven't confirmed it, but I believe that this happens because the instance has [compatibility mode enabled](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/createupdatedomains.html#:~:text=Enable%20compatibility%20mode), which "make[s] OpenSearch report its version as 7.10, which allows certain Elasticsearch OSS clients and plugins that check the version before connecting to continue working with the service."

The solution for this would probably be for OpenSearch to provide a different endpoint that we could check to get the actual version, or to go back to users setting the version itself.

(Alternatively we change the message to just confirm that the version was fetched and not display it. We should double check but I don't think we actually build queries differently for 7.10 and OpenSearch)

iwysiu:
When opensearch is in compatible mode, it returns the version `ElasticSearch 7.10.2` for the version instead of the actual version, so that it can work with elasticseach tooling. This allows the user to correct the auto-detected version. We talked about just hiding the result, but there is a place in the backend where we behave differently for opensearch and elasticsearch 7+ so it does kind of matter.

Fixes #206

### DIFF
--- a/src/configuration/OpenSearchDetails.tsx
+++ b/src/configuration/OpenSearchDetails.tsx
@@ -3,7 +3,7 @@ import { EventsWithValidation, regexValidation, LegacyForms, Button, Alert, Vert
 const { Select, Input, FormField, Switch } = LegacyForms;
 import { Flavor, OpenSearchOptions } from '../types';
 import { DataSourceSettings, SelectableValue } from '@grafana/data';
-import { AVAILABLE_FLAVORS } from './utils';
+import { AVAILABLE_FLAVORS, AVAILABLE_VERSIONS } from './utils';
 import { gte, lt } from 'semver';
 import { OpenSearchDatasource } from 'datasource';
 
@@ -55,13 +55,6 @@ export const OpenSearchDetails = (props: Props) => {
     }
   };
 
-  let versionString = '';
-  if (value.jsonData.flavor && value.jsonData.version) {
-    versionString = `${
-      AVAILABLE_FLAVORS.find((f) => f.value === value.jsonData.flavor)?.label || value.jsonData.flavor
-    } ${value.jsonData.version}`;
-  }
-
   const getServerlessSettings = (event: React.SyntheticEvent<HTMLInputElement, Event>) => {
     // Adds the latest version if it isn't set (query construction requires a version)
     return {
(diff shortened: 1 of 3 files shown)


src/configuration/utils.ts
```diff
@@ -54,7 +54,7 @@ export interface Version {
 
 export const AVAILABLE_VERSIONS: Array<SelectableValue<Version>> = [
   {
-    label: 'OpenSearch 1.0.x',
+    label: 'OpenSearch',
```
iwysiu, line 57: Changed this to just `OpenSearch` so we don't have to deal with keeping up with opensearch releases (they're on 2+ now) but i can add an Opensearch 2.0.x if we'd prefer

src/configuration/OpenSearchDetails.tsx
```diff
@@ -89,7 +82,9 @@ export const OpenSearchDetails = (props: Props) => {
           <VerticalGroup>
             <div>
               The plugin uses the configured version below to construct the queries it sends to the connected OpenSearch
-              instance. If the configured version does not match the instance version, there could be query errors.
+              instance. If the configured version does not match the instance version, there could be query errors. If
+              your OpenSearch instance is configured with &quot;Compatibility Mode&quot;, it will auto-set as
+              &quot;Elasticsearch 7.10.2&quot;. You can correct it with the dropdown menu.
```
njvrzm, line 87: I wonder if this should be a bit more detailed - "you can correct it" doesn't tell me whether or when I _should_ correct it. We don't expect any issues in this case, right? And since the first part of this is about possible problems with version conflicts it might be clearer to have this be on a separate line.

njvrzm (approved):
Looks good - one non-blocking comment.

One thought for another time - it looks like from the response to the version check that we _could_ detect when it's opensearch in compatibility mode: the "tagline" is different, and some other details. It might be nice to show something like "Opensearch (in compatibility mode)" or the like, although I haven't really thought that through 😅

iwysiu:
Closing in favor of using the tagline to autodetect opensearch

iwysiu:
new pr: https://github.com/grafana/opensearch-datasource/pull/419

PR #429, opened 2024-08-02
Chore: Add pre-commit hook

idastambuk:

Adds precommit checks:
- frontend lint with --fix
- backend lint with --fix
- spellcheck

with lefthook (https://github.com/evilmartians/lefthook) which we use in main Grafana repo


njvrzm (approved):
dig it 👍🏻

PR #437, opened 2024-08-23
Add support for PDC

### LINKED ISSUE #431: Add support for using the PDC (Private Data Connection) with this datasource
No

Please can you support the use of the PDC (Private Data Connection) for this datasource. Although it is possible to use the Elasticsearch datasource plugin with its warnings of versions it would be better suited to have this built into the OpenSearch especially as AWS OpenSearch gives you the option to run Opensearch within a private VPC

It is possible to use the Elasticsearch Datasource but it's not ideal

**Additional context**
None required

iwysiu:
Customers have been asking us to support PDC in Opensearch, and now that the backend migration is done, we're finally able to

**Which issue(s) this PR fixes**: #

Fixes #431

You can test it with https://wiki.grafana-ops.net/w/index.php/Engineering/Grafana/Data_Sources/API_servers/Testing_datasources_with_PDC_Locally
I confirmed it worked by setting it up and seeing it work with the proxy on and seeing queries fail with the proxy off.

It turns out that the [implementation instructions](https://docs.google.com/document/d/1IqYANEnTZ30ENRsWGd-oOf0eM44n8ywfytE4U__JEtQ/edit#heading=h.dix66460452b) for implementing this with the HTTP provider were just using the one from the backend sdk that we're already using.

### DIFF
--- a/pkg/opensearch/client/client.go
+++ b/pkg/opensearch/client/client.go
@@ -622,9 +622,6 @@ func (c *baseClientImpl) ExecutePPLQuery(ctx context.Context, r *PPLRequest) (*P
 			},
 		}
 	}
-	if err != nil {
-		return nil, err
-	}
 	return &pr, nil
 }
 
--- a/src/configuration/ConfigEditor.tsx
+++ b/src/configuration/ConfigEditor.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect, useState } from 'react';
-import { DataSourceHttpSettings } from '@grafana/ui';
+import { DataSourceHttpSettings, SecureSocksProxySettings } from '@grafana/ui';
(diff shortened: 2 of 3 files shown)


pkg/opensearch/client/client.go
```diff
@@ -622,9 +622,6 @@ func (c *baseClientImpl) ExecutePPLQuery(ctx context.Context, r *PPLRequest) (*P
 			},
 		}
 	}
-	if err != nil {
```
iwysiu, line 625: this was just a linter thing that was annoying me. err can never be non-nil at this point

kevinwcyu (approved):

iwysiu:
Just to confirm, i rebased it on main locally, and it does run the resource handler queries through the proxy

PR #447, opened 2024-09-02
Bugfix: Update aws/aws-sdk-go to support Pod Identity credentials

idastambuk:
We had a [bug](https://github.com/grafana/opensearch-datasource/issues/435) reported with problems around using Pod Identity for AWS EKS. Looking through the docs it seems like our version of sdk should be updated to support it: https://docs.aws.amazon.com/eks/latest/userguide/pod-id-minimum-sdk.html
Not sure if this is what caused it, but can't hurt to try it

Did a quick sanity check with local open search as well as sigv4 serverless from provisioning and the queries return 200.


njvrzm (approved):
🤞🏻

Issue #434, opened 2024-07-18
Have trace panel display all attributes of an event attached to a span


09jvilla:
**Why is this needed**:

When looking at a span with an associated event in the trace panel, Grafana only shows the event name and event time. It does not show any of the other attributes associated with that event. These other attributes provide important context about the event and aid in the debugging process. 

Here's some screenshots of what this looks like today [using the OpenSearch datasource to view a single trace](https://grafana.com/grafana/plugins/grafana-opensearch-datasource/). 

Below you'll see the complete set of attributes associated with the event, which includes `level`, `code@filepath`, `code@namespace`, `code@lineno`, `offer_ids`, and `target`. (Values have been obfuscated). 

![full-set-of-attributes](https://github.com/user-attachments/assets/6bd2f088-5500-4424-81c3-f86ea1a7fc68)

However, here is how the event is displayed in Grafana. You only see the time and the event name, but none of the attributes associated with that event (e.g., `level`, `offer_ids`). 
![whatsingrafana](https://github.com/user-attachments/assets/49f3aae0-1ee3-4410-8826-cc51eafc1308)

**What would you like to be added**:

The trace panel should show not just the event name and time. It should also show all that event's attributes and the values of those attributes when I expand the event. 

**Who is this feature for?**

Anyone using the traces panel to inspect a trace, and who specifically is looking at spans with events within that trace.

PR #448, opened 2024-09-04
Trace View: Add name and attributes into event object

idastambuk:

- There was [improvement](https://github.com/grafana/grafana/pull/91382) in the UI for Trace View that displays event name more prominently so this adds the "name" as a separate field into the event object
- Improves types (to use Logs and KeyValue types)
- Fixes https://github.com/grafana/opensearch-datasource/issues/434 by adding the event attributes into the fields object

Before (only displaying attribute key values in the title):
Screenshot 2024-09-04 at 19 14 46

After - name displayed at the top + more attributes added to the collapse section:
Screenshot 2024-09-04 at 18 14 57
Screenshot 2024-09-04 at 18 14 20

Fixes https://github.com/grafana/opensearch-datasource/issues/434 

### DIFF
--- a/pkg/opensearch/response_parser.go
+++ b/pkg/opensearch/response_parser.go
@@ -1399,7 +1399,6 @@ func transformTraceEventsToLogs(events []interface{}) ([]Log, []string, error) {
 				}
 				spanEvents = append(spanEvents, Log{Timestamp: timeStamp, Name: eventObj["name"].(string), Fields: eventFields})
 			}
-
 		}
 	}
 	return spanEvents, stackTraces, nil
--- a/pkg/opensearch/response_parser.go
+++ b/pkg/opensearch/response_parser.go
@@ -1387,20 +1387,17 @@ func transformTraceEventsToLogs(events []interface{}) ([]Log, []string, error) {
 			if eventObj, exists := event.(map[string]interface{}); exists {
 				if eventAttributes, exists := eventObj["attributes"].(map[string]interface{}); exists {
 					eventFields = getTraceKeyValuePairs(eventAttributes)
+					// get stack traces if error event
+					errorValue := eventAttributes["error"]
+					if errorValue != nil {
+						stackTraces = append(stackTraces, fmt.Sprintf("%s: %s", eventObj["name"], eventAttributes["error"]))
+					}
 				}
 				timeStamp, err := utils.TimeFieldToMilliseconds(eventObj["time"])
 				if err != nil {
 					return nil, nil, err
 				}
 				spanEvents = append(spanEvents, Log{Timestamp: timeStamp, Name: eventObj["name"].(string), Fields: eventFields})
-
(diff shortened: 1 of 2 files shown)


pkg/opensearch/response_parser.go
```diff
@@ -1360,34 +1370,39 @@ func getErrorFromOpenSearchResponse(response *client.SearchResponse) error {
...
+				spanEvents = append(spanEvents, Log{Timestamp: timeStamp, Name: eventObj["name"].(string), Fields: eventFields})
 
-			// get stack traces if error event
-			attributes, ok := eventObj["attributes"].(map[string]interface{})
-			if ok {
-				errorValue := attributes["error"]
-				if errorValue != nil {
-					stackTraces = append(stackTraces, fmt.Sprintf("%s: %s", eventObj["name"], attributes["error"]))
+				// get stack traces if error event
+				if attributes, exists := eventObj["attributes"].(map[string]interface{}); exists {
```
iwysiu, line 1398: nit: could this be combined with the for loop on line 1387? or do we think its more readable this way

pkg/opensearch/response_parser.go
```diff
@@ -1360,34 +1370,39 @@ func getErrorFromOpenSearchResponse(response *client.SearchResponse) error {
...
+					errorValue := attributes["error"]
+					if errorValue != nil {
+						stackTraces = append(stackTraces, fmt.Sprintf("%s: %s", eventObj["name"], attributes["error"]))
+					}
 				}
 			}
+
```
iwysiu, line 1405:
```suggestion
```

pkg/opensearch/response_parser_test.go
```diff
@@ -2668,14 +2668,21 @@ func TestProcessTraceSpans_creates_correct_data_frame_fields(t *testing.T) {
...
-
-}
-
-type KeyValue struct {
-	Key   string
-	Value any
+	assert.Equal(t, sortedLogs[0].Timestamp, int64(1697615918486))
+	assert.Equal(t, sortedLogs[0].Name, "redis timeout")
+	// assert log attribute values
+	for i, logField := range sortedLogs[0].Fields {
```
iwysiu, line 2675: are we using the key from sortedLogs[0] and the value from sortedLogs[1] for a reason? It's not necessarily wrong (i assume they're the same keys?) but it just seems kind of odd

iwysiu (commented):
Overall looks good! just a couple questions

iwysiu (approved):

pkg/opensearch/response_parser.go
idastambuk: Done!

pkg/opensearch/response_parser_test.go
idastambuk: Definitely an oversight, fixed

PR #449, opened 2024-09-04
Add errorsource to errors

iwysiu:
Adds errorsource to the opensearch datasource.

Fixes https://github.com/grafana/oss-plugin-partnerships/issues/1014

### DIFF
--- a/pkg/opensearch/lucene_handler.go
+++ b/pkg/opensearch/lucene_handler.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/bitly/go-simplejson"
 	"github.com/grafana/grafana-plugin-sdk-go/backend"
+	"github.com/grafana/grafana-plugin-sdk-go/experimental/errorsource"
 	"github.com/grafana/opensearch-datasource/pkg/opensearch/client"
 	"github.com/grafana/opensearch-datasource/pkg/tsdb"
 	"github.com/grafana/opensearch-datasource/pkg/utils"
@@ -279,14 +280,17 @@ func (h *luceneHandler) executeQueries(ctx context.Context) (*backend.QueryDataR
 		return nil, nil
 	}
 
+	response := backend.NewQueryDataResponse()
 	req, err := h.ms.Build()
 	if err != nil {
-		return nil, err
+		return errorsource.AddPluginErrorToResponse(h.queries[0].RefID, response, err), nil
 	}
 
 	res, err := h.client.ExecuteMultisearch(ctx, req)
 	if err != nil {
-		return nil, err
+		backend.Logger.Warn("executeMultisearch", "err", err)
+		// We are returning the error containing the source that was added through errorsource.Middleware
+		return errorsource.AddErrorToResponse(h.queries[0].RefID, response, err), nil
 	}
 
 	rp := newResponseParser(res.Responses, h.queries, res.DebugInfo, h.client.GetConfiguredFields(), h.dsSettings)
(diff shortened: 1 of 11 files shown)


pkg/opensearch/opensearch.go
```diff
@@ -60,25 +61,25 @@ func (ds *OpenSearchDatasource) QueryData(ctx context.Context, req *backend.Quer
...
 	return response, err
 }
 
 // handleServiceMapPrefetch inspects the given request, and, if it wants a serviceMap, creates and
 // calls the Prefetch query to get the services and operations lists that are required for
 // the associated Stats query. It then adds these parameters to the originating query so
-// the Stats query can be created later.
-func handleServiceMapPrefetch(ctx context.Context, osClient client.Client, req *backend.QueryDataRequest) (string, error) {
+// the Stats query can be created later. Returns true if the serviceMap was fetched
+func handleServiceMapPrefetch(ctx context.Context, osClient client.Client, req *backend.QueryDataRequest) *backend.QueryDataResponse {
```
iwysiu, line 78: There wasn't really a better way to handle the fact that `execute` is now returning the error as part of the response and still wrap the other errors

pkg/opensearch/opensearch.go
```diff
@@ -99,41 +102,33 @@ func handleServiceMapPrefetch(ctx context.Context, osClient client.Client, req *
...
+		response := backend.NewQueryDataResponse()
+		err = errorsource.SourceError(source, fmt.Errorf(`Error fetching service map info: %w`, err), true)
+		return errorsource.AddErrorToResponse(refId, response, err)
 	}
-	return nil, err
+	return nil
 }
 
 func wrapError(response *backend.QueryDataResponse, err error) (*backend.QueryDataResponse, error) {
-	var invalidQueryTypeError invalidQueryTypeError
```
iwysiu, line 124: now the error gets wrapped earlier, so we don't need this

pkg/opensearch/lucene_handler.go
```diff
@@ -279,14 +280,16 @@ func (h *luceneHandler) executeQueries(ctx context.Context) (*backend.QueryDataR
...
 	if err != nil {
-		return nil, err
+		return errorsource.AddPluginErrorToResponse(h.queries[0].RefID, response, err), nil
 	}
 
 	res, err := h.client.ExecuteMultisearch(ctx, req)
 	if err != nil {
-		return nil, err
+		// We are returning the error containing the source that was added through errorsource.Middleware
+		return errorsource.AddErrorToResponse(h.queries[0].RefID, response, err), nil
```
nmarrs, line 292: Should we maybe draw out the `h.queries[0].RefID` into a defined reusable var in this case?

pkg/opensearch/opensearch.go
```diff
@@ -60,25 +61,25 @@ func (ds *OpenSearchDatasource) QueryData(ctx context.Context, req *backend.Quer
...
-	if err != nil {
-		return wrapServiceMapPrefetchError(errRefID, err)
+	response := handleServiceMapPrefetch(ctx, osClient, req)
+	if response != nil {
+		return response, nil
 	}
 
 	query := newQueryRequest(osClient, req.Queries, req.PluginContext.DataSourceInstanceSettings)
-	response, err := wrapError(query.execute(ctx))
+	response, err = wrapError(query.execute(ctx))
```
nmarrs, line 70: what is context of removing the `:=` to `=`? (I am a bit of a Go noob btw)

nmarrs: also should we keep this version of `wrapError` around vs using the newer wrap for this error? or in this case are we confident that this is an error owned by us

nmarrs (commented):
nice!! just left some nooby go questions haha

nmarrs (approved):
🚢

njvrzm (approved):
Looks solid! I like the lots of tests, I don't think we've been doing much of that with the errorsource work so far. One optional suggestion.

PR #452, opened 2024-09-13
Use resource handler to get version

iwysiu:
This pr sets up the resource handler (taken from [elasticsearch](https://github.com/grafana/grafana/blob/d1ffcc22d915ab88599c627275cdcf6203e32084/pkg/tsdb/elasticsearch/elasticsearch.go#L190)) and uses it to get the opensearch version

Fixes #

### DIFF
--- a/pkg/opensearch/opensearch.go
+++ b/pkg/opensearch/opensearch.go
@@ -1,9 +1,14 @@
 package opensearch
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"net/http"
+	"net/url"
+	"path"
+	"strings"
 
 	"github.com/bitly/go-simplejson"
 	"github.com/grafana/grafana-plugin-sdk-go/backend"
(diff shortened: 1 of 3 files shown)


pkg/opensearch/opensearch.go
```diff
@@ -166,3 +171,64 @@ func extractParametersFromServiceMapFrames(resp *backend.QueryDataResponse) ([]s
...
+	})
+}
+
+func createOpensearchURL(req *backend.CallResourceRequest, urlStr string) (string, error) {
+	osUrl, err := url.Parse(urlStr)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse data source URL: %s, error: %w", urlStr, err)
+	}
+	osUrl.Path = path.Join(osUrl.Path, req.Path)
+	esUrlString := osUrl.String()
```
njvrzm, line 226:
```suggestion
	osUrlString := osUrl.String()
```

pkg/opensearch/opensearch.go
```diff
@@ -166,3 +171,64 @@ func extractParametersFromServiceMapFrames(resp *backend.QueryDataResponse) ([]s
...
+	// If the request path is empty and the URL does not end with a slash, add a slash to the URL.
+	// This ensures that for version checks executed to the root URL, the URL ends with a slash.
+	// This is helpful, for example, for load balancers that expect URLs to match the pattern /.*.
+	if req.Path == "" && esUrlString[len(esUrlString)-1:] != "/" {
```
njvrzm, line 230:
```suggestion
	if req.Path == "" && osUrlString[len(osUrlString)-1:] != "/" {
```

pkg/opensearch/opensearch.go
```diff
@@ -166,3 +171,64 @@ func extractParametersFromServiceMapFrames(resp *backend.QueryDataResponse) ([]s
...
+		return osUrl.String() + "/", nil
+	}
+	return esUrlString, nil
```
njvrzm, line 233:
```suggestion
	return osUrlString, nil
```

pkg/opensearch/opensearch.go
```diff
@@ -166,3 +171,64 @@ func extractParametersFromServiceMapFrames(resp *backend.QueryDataResponse) ([]s
...
+	if req.Path != "" && !strings.HasSuffix(req.Path, "/_mapping") && req.Path != "_msearch" && req.Path != "_mapping" {
+		return fmt.Errorf("invalid resource URL: %s", req.Path)
+	}
+
+	esUrl, err := createOpensearchURL(req, req.PluginContext.DataSourceInstanceSettings.URL)
+	if err != nil {
+		return err
+	}
+
+	request, err := http.NewRequestWithContext(ctx, req.Method, esUrl, bytes.NewBuffer(req.Body))
```
njvrzm, line 190:
```suggestion
	request, err := http.NewRequestWithContext(ctx, req.Method, osUrl, bytes.NewBuffer(req.Body))
```

pkg/opensearch/opensearch.go
```diff
@@ -166,3 +171,64 @@ func extractParametersFromServiceMapFrames(resp *backend.QueryDataResponse) ([]s
...
+	// allowed paths for resource calls:
+	// - empty string for fetching db version
+	// - /_mapping for fetching index mapping, e.g. requests going to `index/_mapping`
+	// - _msearch for executing getTerms queries
+	// - _mapping for fetching "root" index mappings
+	if req.Path != "" && !strings.HasSuffix(req.Path, "/_mapping") && req.Path != "_msearch" && req.Path != "_mapping" {
+		return fmt.Errorf("invalid resource URL: %s", req.Path)
+	}
+
+	esUrl, err := createOpensearchURL(req, req.PluginContext.DataSourceInstanceSettings.URL)
```
njvrzm, line 185:
```suggestion
	osUrl, err := createOpensearchURL(req, req.PluginContext.DataSourceInstanceSettings.URL)
```

pkg/opensearch/opensearch.go
```diff
@@ -166,3 +171,64 @@ func extractParametersFromServiceMapFrames(resp *backend.QueryDataResponse) ([]s
...
+		return err
+	}
+
+	responseHeaders := map[string][]string{
+		"content-type": {"application/json"},
+	}
+
+	if response.Header.Get("Content-Encoding") != "" {
+		responseHeaders["content-encoding"] = []string{response.Header.Get("Content-Encoding")}
+	}
```
njvrzm, line 211:
```suggestion
	if encoding := response.Header.Get("Content-Encoding"); encoding != "" {
		responseHeaders["content-encoding"] = []string{encoding}
	}
```

njvrzm (changes requested):
Looks good - comments are all minor - and works well when it works, but it fails in a pretty ugly way if the request to the opensearch server fails. Screenshot is what you see after "Get Version and Save" if the server isn't running, or you have the wrong auth, or there's TLS error, etc. Probably fine to just put something generic like "Failed to connect to server", at least for now.
Screenshot 2024-09-17 at 12 21 30

njvrzm:
Just to follow on from the above, there's a different error if you point to a valid http endpoint that's not opensearch. Maybe should use the same generic error here too.

Screenshot 2024-09-17 at 15 30 07

iwysiu:
@njvrzm sorry for the messy history- i accidentally amended onto the last commit instead of making a new one. I made the renaming changes and added the `Failed to connect to server` error. Can you send me the valid endpoint you're using that doesn't return the correct data? I think we originally weren't handling that properly but i can add an error for it

njvrzm:
Ah, sorry, when I said "valid endpoint" I just meant like if you point it to http://www.yahoo.com or the like. A valid HTTP endpoint that isn't an opensearch server.

njvrzm (approved):
Looks good! Thanks for the updates. If you want to repress the json error when visiting a non-opensearch http endpoint feel free, but I don't think it's that important :)

iwysiu:
double checked, it looks like the json error is already being supressed, so i think we're good!

PR #456, opened 2024-09-23
chore: bump grafana-aws-sdk and grafana-plugin-sdk-go

njvrzm:
Adapted from #453, which needed the aws bump, which then entailed the sigv4 auth change.

Closes #453


pkg/opensearch/client/client.go
```diff
@@ -54,7 +55,11 @@ func NewDatasourceHttpClient(ctx context.Context, ds *backend.DataSourceInstance
 		if settings.IsServerless {
 			httpClientOptions.SigV4.Service = "aoss"
 		}
-		httpClientOptions.Middlewares = append(httpClientOptions.Middlewares, sigV4Middleware())
+		authSettings, _ := awsds.ReadAuthSettingsFromContext(ctx)
```
iwysiu, line 58: I think you have to use the deprecated ReadAuthSettings or it doesn't work on older versions of grafana

njvrzm: Hm... we should probably un-deprecate that, in that case. Given how it works under the hood, maybe we should just remove the deprecation and make `ReadAuthSettingsFromContext` private?

idastambuk (approved):

PR #457, opened 2024-09-23
Add dependabot for updating grafana/plugin-sdk-go

idastambuk:
Right now we have dependabot enabled for security updates. This adds dependabot also for non-security updates just to plugin-sdk-go in open search. The purpose of this is to keep the sdk up to date with the latest observability features for api server: https://github.com/grafana/data-sources/issues/64. This is temporary while decide on our general dependabot process. 
(needs to be enabled in settings once this is merged)

Fixes #

~Converted back to draft since Im not sure if this config will conflict with our already enabled security updates for Dependabot~ confirmed it won't

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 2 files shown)


njvrzm (approved):
👍🏻

PR #471, opened 2024-10-07
Backend migration: migrate getTerms

### LINKED ISSUE #467: convert getTerms to backend resource call
draft: https://github.com/grafana/opensearch-datasource/tree/backend-get-terms

iwysiu:
Makes it so that getTerms queries can run through the backend.

Fixes #467 

This is the dashboard I was using to test it. GetTerms is triggered by changing the value for the filter
[opensearch-external-1728336389759.json](https://github.com/user-attachments/files/17284659/opensearch-external-1728336389759.json)

### DIFF
--- a/pkg/opensearch/opensearch.go
+++ b/pkg/opensearch/opensearch.go
@@ -191,6 +191,7 @@ func (ds *OpenSearchDatasource) CallResource(ctx context.Context, req *backend.C
 	if err != nil {
 		return err
 	}
+	request.Header = req.GetHTTPHeaders()
 
 	response, err := ds.HttpClient.Do(request)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


idastambuk (approved):
Looks good! When I tried to test it I couldn't get any term options back though, is this just my setup or is there something wrong with the query (though it looks fine to me)? It's not related to this PR though so this can be merged, I was just wondering if it's a bug we should fix

iwysiu:
> Looks good! When I tried to test it I couldn't get any term options back though, is this just my setup or is there something wrong with the query (though it looks fine to me)? It's not related to this PR though so this can be merged, I was just wondering if it's a bug we should fix

I suspect that the query in question is very un-optimized, lol. It took a while for the query to return data, after i got it running once I think it was cached on opensearch's end and it was fine

PR #477, opened 2024-10-23
Migrate Annotation Queries to run through the backend

### LINKED ISSUE #466: convert AnnotationQuery to backend resource call

iwysiu:
This migrates the annotation queries to run through the backend, which is required for pdc.

Fixes #466 

This copies the [elasticsearch version of this work](https://github.com/grafana/grafana/pull/68075), including the shortcut of running the queries through the resourcehandler instead of having it run through the query endpoint and parsed on the backend, which would add some work. It also doesn't include work to support variables in the annotation queries that elasticsearch has, but we weren't supporting them to begin with.

### DIFF
--- a/src/opensearchDatasource.ts
+++ b/src/opensearchDatasource.ts
@@ -1,5 +1,5 @@
 import _ from 'lodash';
-import { from, merge, of, Observable } from 'rxjs';
+import { from, merge, of, Observable, lastValueFrom } from 'rxjs';
 import { map, tap } from 'rxjs/operators';
 import {
   DataSourceInstanceSettings,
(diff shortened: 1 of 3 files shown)


src/opensearchDatasource.ts
```diff
@@ -302,78 +328,83 @@ export class OpenSearchDatasource extends DataSourceWithBackend<OpenSearchQuery,
...
+      if (timeEndField) {
+        const timeEnd = getFieldFromSource(source, timeEndField);
+        if (timeEnd) {
+          event.timeEnd = toUtc(timeEnd).valueOf();
         }
+      }
 
-        if (typeof event.tags === 'string') {
-          event.tags = event.tags.split(',');
+      // legacy support for title field
```
idastambuk, line 389: Do we know what this title field is, was it ever supported in OpenSearch?

iwysiu: It looks like this was added in the first commit, so possibly not, but I'm going to leave it for now just in case.

PR #485, opened 2024-10-30
Implement backend health check

### LINKED ISSUE #468: CheckHealth call should be a backend call
Currently, we have a frontend call to `_mapping` that checks the health, we should remove it and use the backend `CheckHealth` call and implement that.

iwysiu:
Implements a separate health check in the backend (instead of just using the resource call). I made it call the backend health check if the toggle is enabled because `testDatasource` call returns the least helpful errors possible when the uses the resource call going through the backend. 

Fixes #468

### DIFF
--- a/pkg/opensearch/opensearch.go
+++ b/pkg/opensearch/opensearch.go
@@ -83,6 +83,7 @@ func (ds *OpenSearchDatasource) CheckHealth(ctx context.Context, req *backend.Ch
 		res.Message = fmt.Sprintf("Failed to get index: %s", err)
 		return res, nil
 	}
+	// Same as the frontend check, we generate the time pattern indices for the last six hours
 	indices, err := ip.GetIndices(&backend.TimeRange{From: time.Now().Add(-6 * time.Hour), To: time.Now()})
 	if err != nil || len(indices) == 0 {
 		res.Status = backend.HealthStatusError
@@ -92,6 +93,7 @@ func (ds *OpenSearchDatasource) CheckHealth(ctx context.Context, req *backend.Ch
 
 	var index string
 	var body []byte
+	// We try the indices until one successfully queries
 	for _, indexName := range indices {
 		index = indexName
 		osUrl, err := createOpensearchURL(index+"/_mapping/field/"+url.PathEscape(timeField), req.PluginContext.DataSourceInstanceSettings.URL)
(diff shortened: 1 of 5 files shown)


pkg/opensearch/opensearch.go
```diff
@@ -45,8 +46,122 @@ func NewOpenSearchDatasource(ctx context.Context, settings backend.DataSourceIns
...
 
+	jsonDataStr := req.PluginContext.DataSourceInstanceSettings.JSONData
+	jsonData, err := simplejson.NewJson([]byte(jsonDataStr))
+	if err != nil {
+		res.Status = backend.HealthStatusError
+		res.Message = "Failed to parse settings"
+		return res, nil
+	}
+
+	flavor := jsonData.Get("flavor").MustString("")
```
iwysiu, line 57: Arguably we should do this when we initialize the datasource, but I was trying to avoid changing the logic too much

pkg/opensearch/opensearch.go
```diff
@@ -45,8 +46,124 @@ func NewOpenSearchDatasource(ctx context.Context, settings backend.DataSourceIns
...
+		return res, nil
+	}
+	mapping, ok := jsonData.CheckGet(index)
+	if !ok {
+		res.Status = backend.HealthStatusError
+		res.Message = fmt.Sprintf("Index not found: %s", index)
+		return res, nil
+	}
+
+	timeFieldMapping, ok := mapping.Get("mappings").CheckGet(timeField)
```
iwysiu, line 151: Matches the frontend behavior, which considers it a success if the index is found but the field isn't.

pkg/opensearch/opensearch.go
```diff
@@ -45,8 +46,124 @@ func NewOpenSearchDatasource(ctx context.Context, settings backend.DataSourceIns
...
+		body, err = io.ReadAll(response.Body)
+		if err != nil {
+			res.Status = backend.HealthStatusError
+			res.Message = err.Error()
+			continue
+		}
+
+		if response.StatusCode == 200 {
+			break
+		}
```
njvrzm, line 130: If I understand what you're aiming at here I think we need to add
`res.Status = backend.HealthStatusUnknown` (or `HealthStatusOk`) here before the `break`. Otherwise if the first index we try fails, the check below on line 134 will always be true and we'll return an error, even if a later index succeeds.

njvrzm (approved):
Dig it 👍🏻 Comments optional, except the one about the loop over indexes.

PR #489, opened 2024-11-01
Fix: build each query response separately

kevinwcyu:
When multiple queries are run through the backend flow, it would re-use the same `queryRes` when building the response. This would cause earlier queries' responses to be prepended to subsequent queries' responses. See the next screenshot, where there are 2 queries and both of them have different aliases. In the panel you can see there are 3 series returned and each one is labeled with `alias2`.

before

With this PR, each query's response is built separately. Here is a screenshot of with the fix. There are now 2 series as expected and they and the alias is set correctly now.
after

Fixes 

### DIFF
--- a/pkg/opensearch/response_parser.go
+++ b/pkg/opensearch/response_parser.go
@@ -78,9 +78,6 @@ func (rp *responseParser) parseResponse() (*backend.QueryDataResponse, error) {
 		return result, nil
 	}
 
-	queryRes := backend.DataResponse{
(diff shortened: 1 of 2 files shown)


iwysiu (approved):

Lenormju:
We seem to have encountered this bug while migrating our Grafana to latest version (11.5.2).
Thanks @kevinwcyu for the report and the fix (included in the OpenSearch plugin release 2.21.1).

PR #514, opened 2024-12-03
Add error source, remove some impossible errors

njvrzm:
Some errors without errorsource were causing a high SLO burn rate. I added errorsource there and removed errors from a couple of functions that never returned them.

Fixes https://github.com/grafana/data-sources/issues/230

### DIFF
--- a/pkg/opensearch/client/client.go
+++ b/pkg/opensearch/client/client.go
@@ -6,7 +6,6 @@ import (
 	"crypto/sha256"
 	"encoding/json"
 	"fmt"
-	"github.com/grafana/grafana-plugin-sdk-go/experimental/errorsource"
 	"io"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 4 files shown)


pkg/opensearch/client/client.go
```diff
@@ -108,14 +109,14 @@ func NewClient(ctx context.Context, ds *backend.DataSourceInstanceSettings, http
 
 	version, err := ExtractVersion(jsonData.Get("version"))
 	if err != nil {
-		return nil, fmt.Errorf("version is required, err=%v", err)
+		return nil, errorsource.DownstreamError(fmt.Errorf("version is required, err=%v", err), false)
```
ivanahuckova, line 112: Could you use `backend.DownstreamError()` we are in process of deprecating `experimental/errorsource` and we won't have to fix this later.

pkg/opensearch/client/client.go
```diff
@@ -108,14 +109,14 @@ func NewClient(ctx context.Context, ds *backend.DataSourceInstanceSettings, http
...
 	}
 
 	flavor := jsonData.Get("flavor").MustString(string(OpenSearch))
 
 	timeField, err := jsonData.Get("timeField").String()
 	if err != nil {
-		return nil, fmt.Errorf("time field name is required, err=%v", err)
+		return nil, errorsource.DownstreamError(fmt.Errorf("time field name is required, err=%v", err), false)
```
ivanahuckova, line 119: Same as above

pkg/opensearch/client/client.go
```diff
@@ -133,15 +134,9 @@ func NewClient(ctx context.Context, ds *backend.DataSourceInstanceSettings, http
 		return nil, err
 	}
 
-	indices, err := ip.GetIndices(timeRange)
-	if err != nil {
-		return nil, err
-	}
+	indices := ip.GetIndices(timeRange)
```
ivanahuckova, line 137: 🤩

ivanahuckova (approved):
Awesome! Also great clean up. 👏  Left a comment bellow about using backend.DownstreamError instead of experimenta/errorsource.DownstreamError.

pkg/opensearch/client/client.go
njvrzm: Whoops! Working on autopilot, thanks 😊

PR #523, opened 2024-12-16
fix: parse geohash precision as string instead of int

### LINKED ISSUE #520: Geohash grid precision is poor now
**What happened**:
Metric aggregation with geohash grid results in fixed precision value, regardless of the setting

This can be seen on our public grafana instance : https://grafana.v2.sondehub.org/d/HJgOZLq7k/basic?var-Payload=DG4CG&to=2024-12-14T07:21:39.584Z&from=2024-12-14T01:21:39.584Z&timezone=utc&orgId=1&refresh=1m&viewPanel=panel-7

Image
Image

More precision in our data

Query opensearch wit ha group by geohash grid. Changing the precision does not change the precision in the data returned.

This appears to be a regression as this functionality used to work. 

Looking at git history the only things that stand out to me are:
 - https://github.com/grafana/opensearch-datasource/commit/8c0454b0009d8b199c1826e1b113c9a66fc5e831
 - https://github.com/grafana/opensearch-datasource/commit/20cd33407a9e8d54c409e04a9b692539c75aae62#diff-f57eb533f459265bc8c9e1b89d3834a2afcbf36708ca373e643389c7a1c84573R688

**Environment**:
- Grafana version: v11.4.0 
- OpenSearch version: 7.10.2
- Plugin version: 2.22.1

kevinwcyu:

This fixes an issue where the geohash precision isn't respected. This happens because on the backend it's expecting an int, but the query specifies this as a string. So the precision field always ends up being set to the default value of 3 because of the type mismatch. I think this was probably not noticed until the backend migration completed since the previous frontend flow wouldn't have hit the backend path.

Before (it's always 3 characters returned which doesn't match the precision defined in the query):
before

After:
after

Fixes #520 

In addition to the actual fix, I updated the docker-compose file to include a new required setting for the opensearch nodes. It requires setting `OPENSEARCH_INITIAL_ADMIN_PASSWORD` otherwise the node doesn't start up. I just set it to be `my_%New%_passW0rd!@#` and also updated the `CONTRIBUTING.md` with it.

You can test out the fix by running `yarn server` going to http://localhost:5601, adding sample flight data and running the query shown in the screenshot above.

I also prepared the release in this same PR to hopefully get this out before the freeze.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## 2.22.2
+
+- fix: parse geohash precision as string instead of int in [#523](https://github.com/grafana/opensearch-datasource/pull/523)
+
 ## 2.22.1
+
 - Add error source, remove some impossible errors in [#514](https://github.com/grafana/opensearch-datasource/pull/514)
 
 ## 2.22.0
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,14 +15,14 @@ sysctl -w vm.max_map_count=262144
 ### Add sample data
 
 1. Go to the kibana (http://localhost:5601)
-1. Login with `admin:admin`
+1. Login with `admin:my_%New%_passW0rd!@#`
 1. At the welcome screen click _Add data_ and switch to the _Sample data_ tab.
 1. Import _Sample web logs_ and any other by your choice.
 
 ## Data source configuration
 
 URL: https://localhost:9200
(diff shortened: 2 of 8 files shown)


iwysiu (approved):
Is there a reason the password setting needed to be updated?

kevinwcyu:
> Is there a reason the password setting needed to be updated?

The latest opensearch docker image requires a password to be set otherwise it doesn't start up the opensearch nodes. It doesn't set a default password anymore.

TheSkorm (approved):

PR #545, opened 2025-01-16
Remove openSearchBackendFlowEnabled toggle

idastambuk:

Fixes #


src/opensearchDatasource.ts
```diff
@@ -702,265 +634,25 @@ export class OpenSearchDatasource
...
-    // @ts-ignore-next-line
-    const { openSearchBackendFlowEnabled } = config.featureToggles;
-    const getDbVersionObservable = openSearchBackendFlowEnabled
-      ? this.getResourceRequest('')
-      : this.request('GET', '/');
-    return getDbVersionObservable.then(
+    return this.getResourceRequest('').then(
       (results: any) => {
-        const data = openSearchBackendFlowEnabled ? results : results.data;
+        const data = results;
```
iwysiu, line 655: you can remove this line and use results directly

src/opensearchDatasource.test.ts
```diff
@@ -128,1482 +113,114 @@ describe('OpenSearchDatasource', function (this: any) {
...
-        ],
-      };
-
-      result = await ctx.ds.query(query).toPromise();
-
-      parts = requestOptions.data.split('\n');
-      header = JSON.parse(parts[0]);
-    });
-
-    it('should translate index pattern to current day', () => {
```
iwysiu, line 200: I notice we only kept the `'should not resolve the variable in the original alias field in the query'`, are the other tests just not applicable anymore?

src/opensearchDatasource.test.ts
```diff
@@ -128,1482 +113,114 @@ describe('OpenSearchDatasource', function (this: any) {
...
-    it('should return date fields', async () => {
-      const fieldObjects = await ctx.ds.getFields('date');
-
-      const fields = _.map(fieldObjects, 'text');
-
-      expect(fields).toEqual(['@timestamp_millis']);
-    });
-  });
-
-  describe('When getting fields with backend flow', () => {
```
iwysiu, line 544: I don't think this test should be removed?

src/opensearchDatasource.test.ts
```diff
@@ -128,1482 +113,114 @@ describe('OpenSearchDatasource', function (this: any) {
...
-      const data = mockResource.mock.lastCall[1];
-      const dataArray = data.split('\n');
-      const body = JSON.parse(dataArray[1]);
-      expect(body['aggs']['1']['terms'].size).not.toBe(0);
-    });
-  });
-
-  describe('annotationQuery', () => {
-    describe('results processing', () => {
-      beforeEach(() => {
```
iwysiu, line 905: I also don't think this should have been removed

iwysiu (commented):
The removal looks good, but it seems like tests using the backend flow were also removed from `opensearchDatasource.test.ts`? (I marked a couple of them, but it's worth just going over the code)

src/opensearchDatasource.test.ts
idastambuk: Since the others in the describe block test query building, which we do on the backend now, I don't think so

src/opensearchDatasource.test.ts
idastambuk: Put back these two, got a bit trigger happy

iwysiu (approved):

PR #550, opened 2025-01-21
Make Traces size settable

### LINKED ISSUE #484: Make opensearch traces limit settable
Investigate the opensearch span limits and why traces are currently capped at a limit of 'the first 1000 spans' in the trace

Use Case: Teams can monitor complex systems which often result in traces that include over 1000 spans. What is the reason for the current cap?

iwysiu:
Adds a field to the Lucene Traces query editor for setting the span limits.

Fixes #484 


pkg/opensearch/client/search_request.go
```diff
@@ -618,7 +618,7 @@ func (b *aggBuilderImpl) Stats() AggBuilder {
 }
 
 // TraceList sets the "aggs" object of the query to OpenSearch for the trace list
-func (b *aggBuilderImpl) TraceList() AggBuilder {
+func (b *aggBuilderImpl) TraceList(spanLimit int) AggBuilder {
```
idastambuk, line 621: Then this is technically not just the "span" limit, right? Maybe we could make it more clear in the form too, that we're also setting the trace limit and not just span limit, by renaming the input to `Size`, or something similar? It's a common parameter in OpenSearch queries so I think the users would understand what it means too

iwysiu: renamed!

PR #551, opened 2025-01-22
fix: prevent hidden queries from being used as supplementary queries

svennergr:

Similar to https://github.com/grafana/grafana/pull/99356, hidden queries should not be used for supplementary queries.


idastambuk (approved):

idastambuk:
Thanks @svennergr!

PR #554, opened 2025-01-24
Fix adhoc filters with number fields

iwysiu:
Fixes adhoc filters with number fields

Fixes https://github.com/grafana/grafana/issues/99021

### DIFF
--- a/src/opensearchDatasource.ts
+++ b/src/opensearchDatasource.ts
@@ -194,7 +194,7 @@ export class OpenSearchDatasource
     if (query.hide) {
       return undefined;
     }
-    
+
     let isQuerySuitable = false;
 
     switch (options.type) {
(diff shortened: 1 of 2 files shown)


src/opensearchDatasource.ts
```diff
@@ -1114,9 +1114,10 @@ export class OpenSearchDatasource
 
       const buckets = res.responses[0].aggregations['1'].buckets;
       return _.map(buckets, (bucket) => {
+        const keyString = String(bucket.key);
         return {
-          text: bucket.key_as_string || bucket.key,
-          value: bucket.key,
+          text: bucket.key_as_string || keyString,
+          value: isTagValueQuery ? keyString : bucket.key,
```
iwysiu, line 1120: Technically `MetricFindValue` is supposed to accept numbers as values, but it was logging a console error instead of adding the filter when I tried it, it worked properly when it was a string.

kevinwcyu (approved):

PR #558, opened 2025-01-27
Fix: cast trim edges setting from a string to int

### LINKED ISSUE #544: Trim Edges previously not working in alerts, now no longer works in dashboard panels
**What happened**:
Trim Edges functionality has not been working since v2.16.0 in alerts. The workaround for them is to shift the time from the datetime picker. However, since 2.22.2 it has stopped working in normal visualizations too. Currently, Trim Edges values are not considered and have no effect on a visualization.

The value that is entered to Trim Edges should cut the values from the beginning and end of the values in the panels. (The same behavior as in an Elasticsearch data source.)

1- Create a time series visualization with an OpenSearch data source in a dashboard.
2- Try to enter some values to the Trim Edges input box in the `@timestamp` date histogram aggregation.

This problem previously exists in the alerts but not as critical as exists in visualizations. Every visualization in a Grafana deployment can be affected by this bug.

**Environment**:
- Grafana version: 11.3.2
- OpenSearch version: 2.17.1
- Plugin version: 2.22.2

kevinwcyu:
Trim edges is not working correctly when queries run through the backend flow. There's 2 different issues:

1. The `trimEdges` setting is a string, but is currently getting parsed as an int. This PR casts the setting to an int.
2. The code to delete the data points was skipping data points as it was deleting them so even if the `trimEdges` setting was cast to an int it didn't look like any data points were being trimmed.

Fixes #544 

You can try it out with one of the provisioned data sources. Just set the `Trim Edges` setting to a number that is less than half the total number of data points for the time range you have. With the `Trim Edges` setting, you should see that the number of data points returned is reduced by the 2x the number set for `Trim Edges`. E.g. if there are 11 data points and you set `Trim Edges` to 2, then you should get back 7 data points.

Screenshot 2025-01-27 at 1 18 40 PM

### DIFF
--- a/pkg/opensearch/response_parser.go
+++ b/pkg/opensearch/response_parser.go
@@ -1181,7 +1181,7 @@ func (rp *responseParser) trimDatapoints(frames *data.Frames, target *Query) {
 		return
 	}
 
-	trimEdges, err := histogram.Settings.Get("trimEdges").Int()
+	trimEdges, err := castToInt(histogram.Settings.Get("trimEdges"))
 	if err != nil {
 		return
 	}
@@ -1189,10 +1189,8 @@ func (rp *responseParser) trimDatapoints(frames *data.Frames, target *Query) {
 	for _, f := range *frames {
 		if f.Rows() > trimEdges*2 {
 			for i := 0; i < trimEdges; i++ {
-				f.DeleteRow(i)
-			}
-			for i := f.Rows() - trimEdges; i < f.Rows(); i++ {
-				f.DeleteRow(i)
+				f.DeleteRow(0)
+				f.DeleteRow(f.Rows() - 1)
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


pkg/opensearch/response_parser.go
```diff
@@ -1181,18 +1181,16 @@ func (rp *responseParser) trimDatapoints(frames *data.Frames, target *Query) {
...
 		return
 	}
 
 	for _, f := range *frames {
 		if f.Rows() > trimEdges*2 {
 			for i := 0; i < trimEdges; i++ {
-				f.DeleteRow(i)
-			}
-			for i := f.Rows() - trimEdges; i < f.Rows(); i++ {
-				f.DeleteRow(i)
```
kevinwcyu, line 1195: With the previous logic, this series of data points `| 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 |` and `Trim Edges` set to 2 would be trimmed to `| 1 | 3 | 5 | 6 | 7 | 9 |`.

With the fixed logic, it would be trimmed to `| 3 | 4 | 5 | 6 | 7 |`.

The updated logic just trims from the beginning and end of the series instead of incrementing an index which ends up skipping data points that should be trimmed and not trimming enough data points from the end of the series.

idastambuk (approved):
🎉

PR #565, opened 2025-02-04
Fix logs display when _source is log message field

iwysiu:

Fixes #


iwysiu:
> If I set _source in the `Log message` field, am I supposed to be seeing the raw _source message when log line is collapsed, or did I misunderstand what this is supposed to show? Screenshot 2025-02-05 at 12 05 57

Updated to actually use `_source` as a default log message value

idastambuk:
> > If I set _source in the `Log message` field, am I supposed to be seeing the raw _source message when log line is collapsed, or did I misunderstand what this is supposed to show? Screenshot 2025-02-05 at 12 05 57
> 
> Updated to actually use `_source` as a default log message value

Nice!
Screenshot 2025-02-05 at 17 30 40

idastambuk (approved):

PR #581, opened 2025-03-07
Forward PPL response errors

idastambuk:

It seems like PPL response errors aren't being forwarded to the user, if the Error object is defined in the response. It has this structure: 
Screenshot 2025-03-07 at 10 40 04
Before:
Screenshot 2025-03-07 at 10 47 35
After:
Screenshot 2025-03-07 at 10 46 18

Fixes #


njvrzm (approved):
Tweaked it a bit but 👍🏻

PR #583, opened 2025-03-07
Query Editor: UI improvements and Lucene query type selection improvements

idastambuk:

Our current editor hides Logs, Raw Data and Raw Document under Metrics dropdown, which isn't very intuitive. This refactor the editor to add a few new things: 
1. Adds Lucene Query Selector component, which will update luceneQueryType and metrics values, depending on the selection(logs, raw data, raw document - traces isn't included as a metric in the current query model). N.B. this doesn't change the query model, so even after this Logs and Raw Document queries will have their respective metrics. This is to not to have to deal with migrations, which would be a much bigger feature and should be done separately from this. 
2. Improves display for Alias field - it is only supported with time series queries, i.e. queries with date histogram bucket aggregation
3. Adds an option Size in Logs ,which already exists in the backend, but it seems we have never sent it in the frontend query object. 
Screenshot 2025-03-07 at 15 41 01

Screenshot 2025-03-07 at 15 41 12

Screenshot 2025-03-07 at 15 42 10

Fixes #

### DIFF
--- a/src/components/QueryEditor/LuceneQueryEditor/LuceneQueryEditor.tsx
+++ b/src/components/QueryEditor/LuceneQueryEditor/LuceneQueryEditor.tsx
@@ -29,6 +29,7 @@ export const LuceneQueryEditor = (props: LuceneQueryEditorProps) => {
         <EditorRow>
           <InlineField label="Service Map" tooltip={'Request and display service map data for trace(s)'}>
             <InlineSwitch
+              transparent
               value={props.query.serviceMap || false}
               onChange={(event) => {
                 const newVal = event.currentTarget.checked;
--- a/src/components/QueryEditor/LuceneQueryEditor/LuceneQueryEditor.tsx
+++ b/src/components/QueryEditor/LuceneQueryEditor/LuceneQueryEditor.tsx
@@ -1,21 +1,16 @@
-import { Segment, InlineSegmentGroup, InlineField, InlineSwitch, Input } from '@grafana/ui';
+import { InlineSegmentGroup, InlineField, InlineSwitch, Input } from '@grafana/ui';
 import { useNextId } from 'hooks/useNextId';
 import React from 'react';
 import { LuceneQueryType, OpenSearchQuery } from 'types';
 import { BucketAggregationsEditor } from '../BucketAggregationsEditor';
 import { MetricAggregationsEditor } from '../MetricAggregationsEditor';
 import { QueryEditorRow } from '../QueryEditorRow';
-import { segmentStyles } from '../styles';
+import { LuceneQueryTypeSelector } from './LuceneQueryTypeSelector';
+import { EditorRow, EditorRows } from '@grafana/plugin-ui';
 
 type LuceneQueryEditorProps = {
   query: OpenSearchQuery;
   onChange: (query: OpenSearchQuery) => void;
-};
-const toOption = (queryType: LuceneQueryType) => {
-  return {
-    label: queryType,
-    value: queryType,
-  };
 };
 
 export const LuceneQueryEditor = (props: LuceneQueryEditorProps) => {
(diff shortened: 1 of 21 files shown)


src/components/QueryEditor/LuceneQueryEditor/state.ts
```diff
@@ -0,0 +1,21 @@
+import { type Action, createAction } from '@reduxjs/toolkit';
+import { initQuery } from '../state';
+import { LuceneQueryType, OpenSearchQuery } from 'types';
+
+export const UPDATE_LUCENE_TYPE_AND_METRICS = 'update_lucene_type_and_metrics';
+
+export const updateLuceneTypeAndMetrics = createAction<{
```
idastambuk, line 7: `updateLuceneTypeAndMetrics` updates two reducers (this one and `metricsReducer`) at the same time,. 
 This is because, if I triggered `onChange(query)` from the component and then `changeMetricType`, one would overwrite the other (`changeMetricType` calls `onChange` in the background)

src/components/QueryEditor/index.test.tsx
```diff
@@ -37,22 +39,127 @@ describe('QueryEditorForm', () => {
...
+      };
+
+      render(
+        <QueryEditor query={query} datasource={mockDatasource} onChange={mockOnChange} onRunQuery={mockRunQuery} />
+      );
+
+      expect(screen.queryByLabelText('Alias')).toBeNull();
+    });
+
+    it('Should be shown if last bucket aggregation is Date Histogram', () => {
```
iwysiu, line 145: Should this test have more than 1 bucket arg?

src/components/QueryEditor/MetricAggregationsEditor/state/reducer.ts
```diff
@@ -32,31 +33,11 @@ export const reducer = (state: OpenSearchQuery['metrics'], action: Action): Open
...
-        return {
-          id: metric.id,
-          type: action.payload.type,
-          ...metricAggregationConfig[action.payload.type].defaults,
-        } as MetricAggregation;
-      });
+    return getNewMetrics(state || [], action.payload.id, action.payload.type);
+  }
+  if (updateLuceneTypeAndMetrics.match(action)) {
+    return action.payload.metrics;
```
iwysiu, line 39: nit: from a code organization perspective, it feels a little weird that `changeMetricType` calls `getNewMetrics` here and `updateLuceneTypeAndMetrics` calls it in the `LuceneQueryTypeSelector` when it dispatches the action

iwysiu (approved):
A couple style nits, but I tried it and it looks good!

src/components/QueryEditor/MetricAggregationsEditor/state/reducer.ts
idastambuk: Updated to send metric type and id instead of metrics, so just reducer will call `getNewMetrics`

src/components/QueryEditor/index.test.tsx
idastambuk: Good catch, fixed!

src/components/QueryEditor/LuceneQueryEditor/LuceneQueryEditor.tsx
```diff
@@ -1,88 +1,72 @@
...
         </InlineSegmentGroup>
       </QueryEditorRow>
-      {luceneQueryType === LuceneQueryType.Metric && (
-        <>
-          <MetricAggregationsEditor nextId={nextId} />
-          <BucketAggregationsEditor nextId={nextId} />
-        </>
+      {props.query.luceneQueryType === LuceneQueryType.Traces && (
+        <EditorRow>
+          <InlineField label="Service Map" tooltip={'Request and display service map data for trace(s)'}>
```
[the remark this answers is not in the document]
idastambuk: Makes sense, did it in this one!
Screenshot 2025-03-24 at 19 19 32

PR #585, opened 2025-03-07
Add sample dashboards

kevinwcyu:
I tried to recreate some of the visualizations from the sample dashboards from OpenSearch Dashboards.

- Pull the branch
- Run `yarn server`
- Go to http://localhost:5601/app/home#/tutorial_directory and add the sample data

## Sample ecommerce dashboard

ecommerce

## Sample web traffic logs dashboard

web-logs

## Sample traces dashboard

traces

### DIFF
--- a/cspell.config.json
+++ b/cspell.config.json
@@ -9,7 +9,8 @@
     "go.mod",
     ".gitignore",
     ".config/**",
-    "pkg/opensearch/snapshot_tests/testdata/**"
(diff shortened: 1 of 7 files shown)


idastambuk (approved):

idastambuk:
Not super important, but we talked about how the "avg transactions per day" viz doesn't match the value in OpenSearch. I think just updating the visualization setting to take Calculation => Mean value should fix it.

PR #591, opened 2025-03-14
Bugfix: Fix placeholder text for query string not updating

idastambuk:

The placeholder text for query input field for some reason doesn't change even though we're already passing the correct placeholder prop. I tried adding a key, which forces it to update I guess, and it seems like that fixed it. 

Before: 

Screenshot 2025-03-14 at 12 16 35

After:

Screenshot 2025-03-14 at 12 16 16

Fixes #


njvrzm (approved):
🫡

PR #631, opened 2025-05-08
Add resolution for rxjs

nojaf:

I'm having a bunch of TypeScript type check errors when I run `yarn typecheck`.
All of these were related to a mismatch in rxjs version.
Adding the `resolutions` part solved that for me.

Fixes #

@idastambuk I don't think `yarn typecheck` is part if the CI on https://drone.grafana.net/grafana/opensearch-datasource

Could this be added somehow?

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,5 +106,8 @@
   "engines": {
     "node": ">=20"
   },
+  "resolutions": {
+    "rxjs": "7.8.2"
+  },
   "packageManager": "yarn@1.22.19"
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -3,133 +3,126 @@
 
 
 "@adobe/css-tools@^4.4.0":
-  version "4.4.0"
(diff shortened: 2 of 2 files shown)


idastambuk:
Thanks for this, yes 7.8.1 comes from Grafana packages which will be updated soon-ish but until then we should add the resolution. 
There's some conflicts since we merged some dependabot PRs - can you merge main into this branch, that should fix it. 
Regarding adding it to the workflow, we plan on migrating away from drone very soon so afaik typecheck will be covered with actions after :)

idastambuk:
Hi again @nojaf I think we need to revert the string-width related dependency changes in the yarn.lock (and just keep rxjs), it seems like they're breaking the pipeline 🫠

idastambuk:
I don't think this is a problem anymore, so closing: https://github.com/grafana/opensearch-datasource/actions/runs/16378828553/job/46285301271

PR #633, opened 2025-05-08
Enable live tailing

### LINKED ISSUE #299: Live tailing support for OpenSearch datasource
### Discussed in https://github.com/grafana/opensearch-datasource/discussions/23

<div type='discussions-op-text'>

<sup>Originally posted by **louzadod** November 28, 2021</sup>
Hi.
Consider the possibility of supporting live tailing for OpenSearch datasource as it is implemented for [Loki](https://grafana.com/docs/grafana/latest/datasources/loki/#live-tailing) datasource.

Thanks.</div>

nojaf:

I'm trying to add live tailing / streaming.

Fixes #299


nojaf:
Hello @idastambuk,

This is working for me locally. New logs are found via a polling mechanism. However, there is a difference in the UI: my tailed logs don’t display any colors or details, and I can't see any difference in the responses between a regular call and a WebSocket message.

image

To summarize, I would really appreciate a review from the team to help improve this.

Many thanks!

src/opensearchDatasource.ts
```diff
@@ -647,6 +651,18 @@ export class OpenSearchDatasource
   }
 
   query(request: DataQueryRequest<OpenSearchQuery>): Observable<DataQueryResponse> {
+    console.log('[OpenSearchDatasource] query', request);
+
+    if (request.liveStreaming && !this.isHandlingLiveStreaming) {
```
idastambuk, line 656: Wondering, when is `this.isHandlingLiveStreaming` relevant for the if statement? I added a breakpoint to line 656 and ran the query and I don't see it being anything other than equal to request.liveStreaming at any point `query` is invoked, but I might be missing something

src/opensearchDatasource.ts
```diff
@@ -288,7 +292,7 @@ export class OpenSearchDatasource
       resourceOptions.headers['x-amz-content-sha256'] = await sha256(data);
     }
 
-    return this.postResource(path, data, resourceOptions);
+    return this.postResource(path, JSON.stringify(data), resourceOptions);
```
idastambuk, line 295: What's the reason for this change?

pkg/opensearch/opensearch.go
```diff
@@ -179,12 +224,18 @@ func (ds *OpenSearchDatasource) CheckHealth(ctx context.Context, req *backend.Ch
 // The QueryDataResponse contains a map of RefID to the response for each query, and each response
 // contains Frames ([]*Frame).
 func (ds *OpenSearchDatasource) QueryData(ctx context.Context, req *backend.QueryDataRequest) (*backend.QueryDataResponse, error) {
+	ds.logger.Info("QueryData called", "numQueries", len(req.Queries))
```
idastambuk, line 227: Could we do a review of all the lines where we log stuff, both FE and BE? They're good for debugging, but I think it's better to be conservative since they might be a burden on our cloud logs

idastambuk (commented):
Hi @nojaf, here are some initial comments, will take a look at the rest and the data frame bug separately
Thanks for submitting this!

src/opensearchDatasource.ts
```diff
@@ -662,6 +678,103 @@ export class OpenSearchDatasource
     );
   }
 
+  runLiveQueryThroughBackend(request: DataQueryRequest<OpenSearchQuery>): Observable<DataQueryResponse> {
+    const liveQueries = request.targets.filter((query) => {
```
idastambuk, line 682:
```suggestion
    const logQueries = request.targets.filter((query) => {
```

pkg/opensearch/resource_handlers.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"strings"
+
+	"github.com/grafana/grafana-plugin-sdk-go/backend"
+)
+
+// handleRegisterStreamQuery handles the HTTP request to register a query for a streaming session.
```
idastambuk, line 12: I'm still trying to understand the flow here, can you clarify why this step is necessary? I'm looking at [this guide](https://grafana.com/developers/plugin-tools/tutorials/build-a-streaming-data-source-plugin) to enable streaming in datasource plugins
and I don't see this step to register backend stream, it seems like `getGrafanaLiveSrv().getDataStream({` on the frontend is enough.

pkg/opensearch/lucene_handler.go
```diff
@@ -132,20 +133,6 @@ func processLogsQuery(q *Query, b *client.SearchRequestBuilder, from, to int64,
 		size = defaultLogsSize
 	}
 	b.Size(size)
-
```
idastambuk, line 135: Don't we need this to get the correct data back?

pkg/opensearch/stream_handler.go
```diff
@@ -0,0 +1,168 @@
...
+			osClient, err := client.NewClient(ctx, req.PluginContext.DataSourceInstanceSettings, o.httpClient, &backendQuery.TimeRange)
+			if err != nil {
+				o.logger.Error("RunStream: failed to create OpenSearch client for poll", "refId", refId, "error", err)
+				continue
+			}
+
+			queryExecutor := newQueryRequest(osClient, []backend.DataQuery{backendQuery}, req.PluginContext.DataSourceInstanceSettings)
+			queryDataResponse, err := queryExecutor.execute(ctx)
+
+			if err != nil {
```
idastambuk, line 117: Query response returned contains an .Error field instead of returning an error, so this error wouldn't get logged or passed down to the user

idastambuk:
> Hello @idastambuk,
> 
> This is working for me locally. New logs are found via a polling mechanism. However, there is a difference in the UI: my tailed logs don’t display any colors or details, and I can't see any difference in the responses between a regular call and a WebSocket message.
> 
> image
> 
> To summarize, I would really appreciate a review from the team to help improve this.
> 
> Many thanks!

I'm not sure how to fix this, I tried it in Loki and I see the same issue not being able to expand results: 

Screenshot 2025-06-13 at 14 38 58

src/opensearchDatasource.ts
```diff
@@ -662,6 +678,103 @@ export class OpenSearchDatasource
...
+    }
+
+    const { format, ...streamingQueryPayload } = firstLiveQuery;
+    const finalStreamingQuery = {
+      ...streamingQueryPayload,
+      query: streamingQueryPayload.query || '*',
+      refId: firstLiveQuery.refId,
+    };
+
+    const streamPath = `tail/${firstLiveQuery.refId}`;
```
idastambuk, line 703: I think we can improve the stream path here:
streaming should be reusable - if one user is streaming a query from OpenSearchDS_123 and another user is requesting the same query, this stream should be reused. SO our stream path should reflect this - it can be done same as Loki [here](https://github.com/grafana/grafana/blob/ac832c157e12f6c7c65f74f73d5ea92541eb7d8f/public/app/plugins/datasource/loki/streaming.ts#L22) This means it doesn't need the key either, since the stream path functions as the key later in stream_handler (instead of refid)
Does that make sense?

src/opensearchDatasource.ts
```diff
@@ -662,6 +678,103 @@ export class OpenSearchDatasource
...
+    const liveQueries = request.targets.filter((query) => {
+      return query.metrics?.length === 1 && query.metrics[0].type === 'logs' && !query.hide;
+    });
+
+    if (liveQueries.length === 0) {
+      console.log('[OpenSearch runLiveQueryThroughBackend] No live stream-able queries found.');
+      return of({ data: [], state: LoadingState.Done, key: request.requestId });
+    }
+
+    const firstLiveQuery = liveQueries[0];
```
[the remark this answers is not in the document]
nojaf: Yes, good catch, will remove it.

pkg/opensearch/opensearch.go
nojaf: I'm happy to remove or update the levels afterward. The main reason for adding them was to gauge how far along we are in the pipeline.

pkg/opensearch/lucene_handler.go
nojaf: I believe I started messing with this because I never saw my entire data.
I keep seeing only this:
image

pkg/opensearch/stream_handler.go
```diff
@@ -0,0 +1,170 @@
...
+			}
+
+			if !backendQuery.TimeRange.To.After(backendQuery.TimeRange.From) {
+				o.logger.Debug("RunStream: 'To' time is not after 'From' time, skipping poll to avoid empty range", "from", backendQuery.TimeRange.From, "to", backendQuery.TimeRange.To)
+				continue
+			}
+
+			o.logger.Info("RunStream: Polling OpenSearch", "refId", refId, "from", backendQuery.TimeRange.From, "to", backendQuery.TimeRange.To, "duration", backendQuery.TimeRange.To.Sub(backendQuery.TimeRange.From))
+
+			osClient, err := client.NewClient(ctx, req.PluginContext.DataSourceInstanceSettings, o.httpClient, &backendQuery.TimeRange)
```
idastambuk, line 97: I think we we should be able to just run 
```
queryDataResponse, err :=  o.QueryData(ctx, &backend.QueryDataRequest{
				Queries:  []backend.DataQuery{backendQuery},
				PluginContext: req.PluginContext,	
				Headers:  req.Headers,
			})
```
instead of this and the following lines, since we have all the data to construct the data query request.

src/opensearchDatasource.ts
```diff
@@ -662,6 +676,100 @@ export class OpenSearchDatasource
...
 
+  async getLiveStreamKey(query: OpenSearchQuery): Promise<string> {
+    const str = JSON.stringify({ query: query.query });
+    const msgUint8 = new TextEncoder().encode(str); // encode as (utf-8) Uint8Array
+    const hashBuffer = await crypto.subtle.digest('SHA-1', msgUint8); // hash the message
+    const hashArray = Array.from(new Uint8Array(hashBuffer.slice(0, 8))); // first 8 bytes
+    return `${query.datasource?.uid}/${hashArray.map((b) => b.toString(16).padStart(2, '0')).join('')}/${config.bootData.user.orgId}`;
+  }
+
+  runLiveQueryThroughBackend(request: DataQueryRequest<OpenSearchQuery>): Observable<DataQueryResponse> {
```
idastambuk, line 687: We could also probably extract this function to a separate file to keep the datasource class cleaner

idastambuk:
> Hello @idastambuk,
> 
> This is working for me locally. New logs are found via a polling mechanism. However, there is a difference in the UI: my tailed logs don’t display any colors or details, and I can't see any difference in the responses between a regular call and a WebSocket message.
> 
> image
> 
> To summarize, I would really appreciate a review from the team to help improve this.
> 
> Many thanks!

So I think this is just the feature and not a bug, it renders a basic table: [LiveLogs.tsx#L146](https://github.com/grafana/grafana/blob/566460f4fed7bc3388b937bd10708edbb854732f/public/app/features/explore/Logs/LiveLogs.tsx#L146)
and only when it stops streaming does it actually display the logs visualization with all the extra info: [LogsContainer.tsx#L329](https://github.com/grafana/grafana/blob/566460f4fed7bc3388b937bd10708edbb854732f/public/app/features/explore/Logs/LogsContainer.tsx#L329) Not sure why it was implemented like that, I asked the team that did it if it could be improved

Issue #578, opened 2025-02-28
PPL queries are failing


VermaPriyanka:
**What happened**: 
All PPL queries are failing, while lucene queries are working as expected.
Running the PPL query directly on Opensearch dashboard works fine. Couldn't find anything recorded in the error logs on OpenSearch side.Tried using different browser, but issue persists.

Error received: 
unexpected status code: 400
(Trace ID: 44348c3d-7581-4735-9309-08f63394c125)

**What you expected to happen**:
PPL query to work as expected, similar to how it returns results in OpenSearch Dashboards.

**How to reproduce it (as minimally and precisely as possible)**:
Run any PPL query for an OpenSearch data source. 
Example: source=eks-webstg-v1-*| where kubernetes.namespace_name='s3-dataset'

**Anything else we need to know?**:

**Environment**:
- Grafana version:
- OpenSearch version: 
- Plugin version: 2.23.1

iwysiu:
did it work in a previous version?

iwysiu:
I tried it locally and the query worked. Do you have more information about their Opensearch and their datasource configuration? Does it work when they hit save and test?

adiospeds:
Hi @iwysiu - I am seeing the same issue - 

My Opensearch version is 2.17 (AWS managed Opensearch) and Grafana version is 10.4 (AWS managed Grafana) and Opensearch datasource plugin version is 2.23.1

Also previously I was using Opensearch datasource plugin version is 2.22.1 but still facing same issue, so I upgraded to  2.23.1 but there was no success. As mentioned above the lucene queries work fine and show a graph for the index however as soon as I switch to PPL it starts throwing 

Error received:
unexpected status code: 400
(trace ID follows)

njvrzm:
We've reproduced the issue now in Grafana Cloud as well. No information yet about the cause but we'll dig into it. Thanks for the reports!

idastambuk:
Hi again @VermaPriyanka and @adiospeds, we released v2.32.2 of the plugin with improved error messaging, since `unexpected status code: 400` didn't help debugging it. Can you update the plugin and run your queries once again and see if you're getting a more meaningful message? Thanks!

adiospeds:
I will test this today and get back on how it goes.

adiospeds:
[...]

PR #635, opened 2025-05-11
Remove explicit timestamp conversions

### LINKED ISSUE #632: PPL queries use timestamp() resulting in scripted plan
**What happened**:

PPL queries use `timestamp()` to construct timestamp objects when [searching](https://github.com/grafana/opensearch-datasource/blob/main/pkg/opensearch/client/ppl_request.go#L32), which results in a scripted plan. This is very slow compared to allowing OpenSearch to interpret the raw timestamp directly.

For example executing this query plan takes ~10 seconds on our data set, as it uses a script to filter:

```
POST /_plugins/_ppl/_explain
{
  "query": "index=myindex | where `@timestamp` >= timestamp('2025-05-07 15:04:05') and `@timestamp` <= timestamp('2025-05-07 15:38:30')"
}
{
  "root": {
    "name": "ProjectOperator",
    "description": {
      "fields": "[cloud, http, @timestamp, event, log]"
    },
    "children": [
      {
        "name": "OpenSearchIndexScan",
        "description": {
          "request": """OpenSearchQueryRequest(indexName=myindex, sourceBuilder={"from":0,"size":200,"timeout":"1m","query":{"bool":{"filter":[{"script":{"script":{"source":"<elided>","lang":"opensearch_query_expression"},"boost":1.0}},{"script":{"script":{"source":"<elided>","lang":"opensearch_query_expression"},"boost":1.0}}],"adjust_pure_negative":true,"boost":1.0}},"_source":{"includes":["log","@timestamp","http","cloud","event"],"excludes":[]},"sort":[{"_doc":{"order":"asc"}}]}, needClean=true, searchDone=false, pitId=null, cursorKeepAlive=null, searchAfter=null, searchResponse=null)"""
        },
        "children": []
      }
    ]
  }
}
```

whereas this one without `timestamp()` takes  ~200ms as it can use a range query directly:

```
POST /_plugins/_ppl/_explain
{
  "query": "index=myindex | where `@timestamp` >= '2025-05-07 15:04:05' and `@timestamp` <= '2025-05-07 15:38:30'"
}
{
  "root": {
    "name": "ProjectOperator",
    "description": {
      "fields": "[cloud, http, @timestamp, event, log]"
    },
    "children": [
      {
        "name": "OpenSearchIndexScan",
        "description": {
          "request": """OpenSearchQueryRequest(indexName=myindex, sourceBuilder={"from":0,"size":200,"timeout":"1m","query":{"bool":{"filter":[{"range":{"@timestamp":{"from":1746630245000,"to":null,"include_lower":true,"include_upper":true,"boost":1.0}}},{"range":{"@timestamp":{"from":null,"to":1746632310000,"include_lower":true,"include_upper":true,"boost":1.0}}}],"adjust_pure_negative":true,"boost":1.0}},"_source":{"includes":["@timestamp","event","log","cloud","http"],"excludes":[]},"sort":[{"_doc":{"order":"asc"}}]}, needClean=true, searchDone=false, pitId=null, cursorKeepAlive=null, searchAfter=null, searchResponse=null)"""
        },
        "children": []
      }
    ]
  }
}
```

The plan should avoid expensive queries where possible.

See above.

**Environment**:
- Grafana version: Grafana v11.6.0+security-01 (bb93ae3e12)
- OpenSearch version: AWS Serverless
- Plugin version: 2.25.0

jonnangle:

OpenSearch natively infers the date format used by the plugin (`YYYY-MM-DD HH:MM:SS`) as a timestamp literal, so it seems unnecessary to have an explicit conversion using `timestamp()`. Removing it allows OpenSearch to use a range filter instead of a script in the query plan, which makes a significant difference to performance (see detail in #632)

Fixes #632 

I've replaced the `timestamp()` calls everywhere but I'm not 100% sure whether the frontend code is used or not (the timestamp values applied via filters appear to be longs?) - would appreciate your insights :)

### DIFF
--- a/pkg/opensearch/client/client_test.go
+++ b/pkg/opensearch/client/client_test.go
@@ -190,7 +190,7 @@ func TestClient(t *testing.T) {
 					assert.NoError(t, err)
 
 					t.Run("and replace index pattern with wildcard", func(t *testing.T) {
-						assert.Equal(t, "source = metrics-* | where `@timestamp` >= timestamp('$timeFrom') and `@timestamp` <= timestamp('$timeTo')", jBody.Get("query").MustString())
+						assert.Equal(t, "source = metrics-* | where `@timestamp` >= '$timeFrom' and `@timestamp` <= '$timeTo'", jBody.Get("query").MustString())
 					})
 				})
 				t.Run("Should parse response", func(t *testing.T) {
@@ -239,7 +239,7 @@ func TestClient(t *testing.T) {
 					assert.NoError(t, err)
 
 					t.Run("and replace index pattern with wildcard", func(t *testing.T) {
-						assert.Equal(t, "source = metrics-* | where `@timestamp` >= timestamp('$timeFrom') and `@timestamp` <= timestamp('$timeTo')", jBody.Get("query").MustString())
+						assert.Equal(t, "source = metrics-* | where `@timestamp` >= '$timeFrom' and `@timestamp` <= '$timeTo'", jBody.Get("query").MustString())
 					})
 				})
 				t.Run("Should parse response", func(t *testing.T) {
--- a/pkg/opensearch/client/ppl_request.go
+++ b/pkg/opensearch/client/ppl_request.go
@@ -29,7 +29,7 @@ func (b *PPLRequestBuilder) Build() (*PPLRequest, error) {
 // AddPPLQueryString adds a new PPL query string with time range filter
 func (b *PPLRequestBuilder) AddPPLQueryString(timeField, to, from, querystring string) *PPLRequestBuilder {
 	var res []string
-	timeFilter := fmt.Sprintf(" where `%s` >= timestamp('%s') and `%s` <= timestamp('%s')", timeField, from, timeField, to)
+	timeFilter := fmt.Sprintf(" where `%s` >= '%s' and `%s` <= '%s'", timeField, from, timeField, to)
 
 	trimmedQuerystring := strings.TrimSpace(querystring)
 	// Sets a default query if the query string is empty
--- a/pkg/opensearch/client/ppl_request_test.go
+++ b/pkg/opensearch/client/ppl_request_test.go
@@ -37,7 +37,7 @@ func TestPPLRequest(t *testing.T) {
 
 					t.Run("Should have query string filter", func(t *testing.T) {
 						f := pr.Query
-						assert.Equal(t, "source = default_index | where `@timestamp` >= timestamp('$timeFrom') and `@timestamp` <= timestamp('$timeTo')", f)
+						assert.Equal(t, "source = default_index | where `@timestamp` >= '$timeFrom' and `@timestamp` <= '$timeTo'", f)
 					})
 
 					t.Run("When marshal to JSON should generate correct json", func(t *testing.T) {
@@ -45,7 +45,7 @@ func TestPPLRequest(t *testing.T) {
 						assert.NoError(t, err)
 						json, err := simplejson.NewJson(body)
 						assert.NoError(t, err)
-						assert.Equal(t, "source = default_index | where `@timestamp` >= timestamp('$timeFrom') and `@timestamp` <= timestamp('$timeTo')", json.Get("query").Interface())
+						assert.Equal(t, "source = default_index | where `@timestamp` >= '$timeFrom' and `@timestamp` <= '$timeTo'", json.Get("query").Interface())
 					})
 				})
 			})
@@ -58,7 +58,7 @@ func TestPPLRequest(t *testing.T) {
 
 					t.Run("Should have query string filter", func(t *testing.T) {
 						f := pr.Query
-						assert.Equal(t, "source = index | where `@timestamp` >= timestamp('$timeFrom') and `@timestamp` <= timestamp('$timeTo') | fields test", f)
+						assert.Equal(t, "source = index | where `@timestamp` >= '$timeFrom' and `@timestamp` <= '$timeTo' | fields test", f)
 					})
 
 					t.Run("When marshal to JSON should generate correct json", func(t *testing.T) {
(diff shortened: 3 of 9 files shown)


idastambuk:
Hi @jonnangle,  thanks for submitting this! However, I can't seem to find docs that state that 
> OpenSearch natively infers the date format used by the plugin (YYYY-MM-DD HH:MM:SS) as a timestamp literal
in boolean expression for timestamp - can you link it here? 

Im wondering if, instead of using `timestamp() function`, we can convert our date ranges to epoch millis in BE and THEN add them to the string, since it seems like they're stored like that internally in OS, so there isn't a risk with incorrect parsing: https://docs.opensearch.org/docs/latest/field-types/supported-field-types/date/?

jonnangle:
hi @idastambuk! This is the reference I found for the PPL date format

https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/general/datatypes.rst#timestamp

I found that if I tried to use any other format it would return this error

https://github.com/opensearch-project/sql/blob/main/core/src/main/java/org/opensearch/sql/data/model/ExprTimestampValue.java#L40

I do get OpenSearch parse failures if I try to use a date format that doesn't support `epoch_millis`. For example, if I use this:

```
        "my_date": {
          "type": "date",
          "format": "MM/dd/yyyy HH:mm:ss"
        }
```

then the PPL query `| where my_date >= '2019-01-01 14:00:23'` gets me this:

```
OpenSearchParseException[failed to parse date field [1546351223000] with format [MM/dd/yyyy HH:mm:ss]: [failed to parse date field [1546351223000] with format [MM/dd/yyyy HH:mm:ss]
```

So it seems to me that PPL expects to be able to convert its preferred format to epoch_millis internally and have that be supported by the mapping. Performing an explicit conversion by wrapping in `timestamp()` certainly provides support for mappings that don't allow for epoch_millis, but those seem like an edge case when you look at the cost of filtering via scripts on every query.

Let me know if this doesn't make sense - I'm new to OpenSearch so am very much still learning!

idastambuk:
Hi again @jonnangle and thanks for clarifying. Yeah, that makes sense. 
However, I'm just evaluating another bug report where a few users use non-YYYY-MM-DD HH:MM:SS time formats ([example](https://github.com/grafana/opensearch-datasource/issues/578#issuecomment-2745029300)), meaning this change wouldn't be compatible with their queries. 
It still seems to me that the least-bad approach is to wrap it in `timestamp()` to ensure compatibility with any possible timestamp formats, though that's obviously not ideal from a performance standpoint

There are a few other options we could do. 
- Option 1: Not adding `where timestamp..` in the plugin, but leaving to whoever's writing the query to add it (with or without timestamp()). Unfortunately, this would make it backwards incompatible, since there are a lot of stored queries without this filter
- Option 2: implement a solution where we look at the format in the index mapping resource call (which we don't do for PPL queries right now) and if there's a `format` field defined, send the `where` clause in that format. If there isn't one defined, wrap it in timestamp(). This would obviously be a bigger workload than just this PR.
I'll talk to the team and get back to you how we want to pursue this. Of course, any additional ideas are welcome!

jonnangle:
> Hi again @jonnangle and thanks for clarifying. Yeah, that makes sense. However, I'm just evaluating another bug report where a few users use non-YYYY-MM-DD HH:MM:SS time formats ([example](https://github.com/grafana/opensearch-datasource/issues/578#issuecomment-2745029300)),

ah ok, interesting! `unable to parse time field` seems to come from [here](https://github.com/grafana/opensearch-datasource/blob/main/pkg/opensearch/ppl_response_parser.go#L194-L197) which suggests that the query succeeded, but the `value` from the PPL response that's supposed to be a timestamp is failing the `string` type assertion. Are we certain that there are no mapping issues in play with this?

Just wanted to mention as well - I noticed that the OpenSearchParseException that I mentioned in my last comment appears to be version-dependent. I get the parse error on AWS Serverless, but if I perform the same test when running locally using this version:

```
  "version" : {
    "distribution" : "opensearch",
    "number" : "2.19.1",
    "build_type" : "tar",
    "build_hash" : "2e4741fb45d1b150aaeeadf66d41445b23ff5982",
    "build_date" : "2025-02-27T01:22:24.665339607Z",
```

then PPL does create a suitable `range` query plan for the time format in use (`"MM/dd/YYYY HH:mm:ss"`)

```
POST /_plugins/_ppl/_explain
{
  "query": "index=testindex | where my_date < '2024-05-13 09:00:00'"
}

{
  "root": {
    "name": "ProjectOperator",
    "description": {
      "fields": "[my_date]"
    },
    "children": [
      {
        "name": "OpenSearchIndexScan",
        "description": {
          "request": """OpenSearchQueryRequest(indexName=testindex, sourceBuilder={"from":0,"size":10000,"timeout":"1m","query":{"range":{"my_date":{"from":null,"to":"05/13/2024 09:00:00","include_lower":true,"include_upper":false,"boost":1.0}}},"_source":{"includes":["my_date"],"excludes":[]},"sort":[{"_doc":{"order":"asc"}}]}, needClean=true, searchDone=false, pitId=null, cursorKeepAlive=null, searchAfter=null, searchResponse=null)"""
        },
        "children": []
      }
    ]
  }
}
```

so it seems like there's ongoing work to improve support for implicit conversions between PPL and OpenSearch in this area.

I fully understand the desire to support as many timestamp formats as possible, but unfortunately wrapping in `timestamp()` will make PPL completely unusable for us simply because of the scripted query plans. In our log index of ~25m documents, we'll just get timeouts if we try to look at the past hour in Explore. (AWS Serverless will valiantly try to scale up to cope, but that's not a workable solution when Lucene can fetch the past hour of events in < 1 second by using the index :)

jonnangle:
Also wanted to mention this, in terms of prior art - if I look at my test index via Observability Logs Explorer in OpenSearch Dashboards, it uses the `YYYY-MM-DD HH:MM:SS` format when building a range query for the past 15 minutes - so it feels like this could be the right track for Grafana as well

image

jonnangle:
Thanks for the update @idastambuk!  Please let me know if I can help in any way - our OpenSearch project will be running for a while yet, so we've got some space and time to try out ideas at scale.

jonnangle:
> Also do you mind opening an issue in https://github.com/opensearch-project/sql? Theoretically `TIMESTAMP()` shouldn't cause perf problems

Thanks @joshuali925, I've raised an [issue](https://github.com/opensearch-project/sql/issues/3842).

jonnangle:
Hi @idastambuk - yes that sounds like a plan 👍 We'll probably just run a patched version of the datasource for now to tide us over until the fix makes it into OpenSearch - hopefully it won't be too long. Thanks so much for all your help with this!

PR #636, opened 2025-05-12
Fix permissions for get-vault-secrets

idastambuk:

Fixes #

### DIFF
--- a/.github/workflows/pr-commands.yml
+++ b/.github/workflows/pr-commands.yml
@@ -11,6 +11,8 @@ jobs:
   main:
     permissions:
       contents: read
+      # The "id-token: write" permission is required by "get-vault-secrets" action. We request it here
+      # instead of the whole workflow, in order to reduce the scope of the permission.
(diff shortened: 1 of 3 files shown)


.github/workflows/pr-commands.yml
```diff
@@ -11,6 +11,8 @@ jobs:
   main:
     permissions:
       contents: read
+      # The "id-token: write" permission is required by "get-vault-secrets" action. We request it here
```
njvrzm, line 14:
```suggestion
      # The "pull-requests: write" permission is required by "get-vault-secrets" action. We request it here
```
(I'm assuming the yaml is correct and the comment is not)

njvrzm (approved):
👍🏻

.github/workflows/pr-commands.yml
idastambuk: Good catch, I forgot to add the actual permission for token 🤦🏻‍♀️

PR #639, opened 2025-05-13
Add execution hint support to terms aggregation

enawar-knowunity:
- Updated TermsAggregation model to include an optional execution_hint field.
- Enhanced lucene_handler to handle execution_hint from bucket aggregation settings.
- Modified query_request_test to validate execution_hint functionality.
- Updated BucketAggregationsEditor to include execution hint selection in the UI.

Fixes #

### DIFF
--- a/pkg/opensearch/client/models.go
+++ b/pkg/opensearch/client/models.go
@@ -341,11 +341,12 @@ type BucketScriptAggregation struct {
 
 // TermsAggregation represents a terms aggregation
 type TermsAggregation struct {
-	Field       string                 `json:"field"`
-	Size        int                    `json:"size"`
-	Order       map[string]interface{} `json:"order,omitempty"`
-	MinDocCount *int                   `json:"min_doc_count,omitempty"`
-	Missing     *string                `json:"missing,omitempty"`
+	Field         string                 `json:"field"`
+	Size          int                    `json:"size"`
+	Order         map[string]interface{} `json:"order,omitempty"`
+	MinDocCount   *int                   `json:"min_doc_count,omitempty"`
+	Missing       *string                `json:"missing,omitempty"`
+	ExecutionHint *string                `json:"execution_hint,omitempty"`
 }
 
 // ExtendedBounds represents extended bounds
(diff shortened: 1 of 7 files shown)


idastambuk:
Hi @enawar-knowunity thanks a lot for submitting this! Im trying to find documentation in opensearch about this execution_hint setting and specifically it with terms aggregation, but I can't seem to find anything but this: https://docs.opensearch.org/docs/latest/aggregations/metric/cardinality/#:~:text=Configuring%20aggregation%20execution,Uses%20ordinals%20of%20the%20field.

Can you point me to the docs you used for this?

idastambuk:
> Hi @idastambuk thank you for your message! So the execution_hint for bucket aggregation is not directly mentioned in the Opensearch docs, I think the one you linked is rather about cardinality. However, it is:
> 
> 1. mentioned in the [docs](https://www.elastic.co/docs/reference/aggregations/search-aggregations-bucket-terms-aggregation#search-aggregations-bucket-terms-aggregation-execution-hint) of Elasticsearch
> 2. mentioned in the implementation files of the aggregation operation in opensearch ([link](https://github.com/opensearch-project/OpenSearch/blob/bc3778a29c508cb54407dac0bf5a38313e43956c/server/src/main/java/org/opensearch/search/aggregations/bucket/terms/SignificantTermsAggregatorFactory.java#L414))
> 
> While testing it locally, it heavily optimized some of our queries in opensearch running from 2m to 4s.

Ah, I guess it's a case of opensearch docs not being up to date, all clear now!

enawar-knowunity:
@idastambuk thank you for the suggestion regarding the tooltip and the default value, I've added both, could you take another look?

idastambuk:
Hi again @enawar-knowunity, I'm not sure we should force a default value on the hint, since the user might just want to let open search internally decide on the hint depending on the field type: 
> global_ordinals is the default option for keyword field

but 

>By default, map is only used when running an aggregation on scripts, since they don’t have ordinals.

enawar-knowunity:
@idastambuk fair point, I removed the enforced default and rephrased the tooltip message

idastambuk (approved):

enawar-knowunity:
@idastambuk perfect, thank you so much! Do you know when is the next release going to be? : )

idastambuk:
> @idastambuk perfect, thank you so much! Do you know when is the next release going to be? : )

Just released it: https://github.com/grafana/opensearch-datasource/releases/tag/v2.27.0

PR #643, opened 2025-05-13
Fix: Add support for Elasticsearch 6.8 by handling hits.total as number

ranyhb:
## What
- This PR adds support for Elasticsearch 6.8 by implementing a custom  method for the  struct to handle both number format (used in Elasticsearch 6.8) and object format (used in newer versions).

## Why
- In Elasticsearch 6.8, the  field is returned as a number instead of an object with  and  fields.
- This was causing the error: 'error while Decoding to MultiSearchResponse: json: cannot unmarshal number into Go struct field SearchResponseHits.responses.hits.total of type client.SearchResponseHitsTotal'

## Testing
- Added tests to verify the behavior works with both formats.
- Successfully builds and runs.

Fixes the bug noted in CHANGELOG.md under 2.26.0.


idastambuk:
Hi @ranyhb good catch, and thanks a lot for the contribution! 
This looks good, but can you please remove the changes to the changelog and package.json? We're preparing the same release and we'd like a separate commit for the release that has other commits as well. Thanks!

ranyhb:
@idastambuk 
Done!

PR #657, opened 2025-06-02
Skipping connectivity checks for cross-cluster search

### LINKED ISSUE #610: Support cross cluster queries
OpenSearch data source plugin doesn't allow querying remote indexes when using OpenSearch cross cluster search feature - when trying to use remote index, like `remote_cluster:index-name`, I get 404 error in UI.

I would like OpenSearch data source plugin to be able to retrieve data from remote cluster(s).

ElasticSearch plugin supports it - https://github.com/grafana/grafana/issues/72353. I've succeeded to use it with my OpenSearch cluster but I don't think it's much future-proof.

**Additional context**
Screenshot with index name configuration and error:
Image

i-kata:
Tests:

Zrzut ekranu 2025-06-2 o 21 22 12

Zrzut ekranu 2025-06-2 o 21 23 19

Skips connectivity checks when cross-cluster search is detected. This will allow users to use cross-cluster search feature with a specific remote connection (not only wildcard as it is now). Currently if a user inputs "index,*:index" as Index name it will go through and logs from local and remote cluster will be fetched, but if a user specifies the name of the remote connection - index,remote_connection:index - a 404 will be thrown. This PR will fix it by skipping connectivity checks only when a cross-cluster search is detected.

Fixes #610

### DIFF
--- a/src/opensearchDatasource.ts
+++ b/src/opensearchDatasource.ts
@@ -540,6 +540,13 @@ export class OpenSearchDatasource
     // @ts-ignore-next-line
     // TODO: run through backend health check
 
+    if (this.index && this.index.includes(':')) {
+      return Promise.resolve({
+        status: 'success',
+        message: 'Datasource test skipped - cross-cluster search detected',
+      });
+    }
+
     if (!this.flavor || !valid(this.version)) {
       return Promise.reject({
         status: 'error',


idastambuk:
hi @i-kata, I just left a comment in the original ticket:https://github.com/grafana/opensearch-datasource/issues/610#issuecomment-3073600710

Rather than disabling the health check, we should try to fix it by replacing the API to one that supports cross-cluster setup. This is in our backlog, but if you would like to try implementing it, we're happy to support you with that, let us know!

PR #693, opened 2025-07-15
Fix: process percentile metric queries to display values on the frontend

### LINKED ISSUE #587: Explore and dashboards: result of Percentiles/Extended stats are not shown
### What happened?

Columns percentiles and extended stats are not been split in table mode.
Percentiles always show null value, while Extended Stats keep the last value in response

I catch the request inside the logs of OpenSearch and I was able to replay it and get the good result

Image

### What did you expect to happen?

Columns could be split in multiple columns as fields as they contains

### Did this work before?

I don't know.. (but it works fine with elastic datasource)

### How do we reproduce it?

1. Go to Explore metrics
2. Request integer metric with percentiles
3. Group by a field
4. Table display only one column by metric

### Is the bug inside a dashboard panel?

_No response_

### Environment (with versions)?

Grafana: v11.5.2
Browser: Firefox 135.0.1

### Grafana platform?

Docker

### Datasource(s)?

OpenSearch

kevinwcyu:

- Fixes an issue where percentile metric query types were not being processed. `processAggregationDocs` did not have a case for the `percentilesType` type so it was not processing it correctly.
- Fixes an issue where extended stats metric query types only returned 1 value, if multiple were specified in the query (e.g. if you selected `min` and `max`, it would only return `max`. The break statement caused the processing to only process the first extended stat metric.
- Fixes an issue where the extended stats label only returned `Extended Stats` instead of `Extended Stats {stat name}` e.g. if `min` was one of the selected stats it should say `Extended Stats Min` in the label.

Fixes #587 

You can test this by:
1. Build the data source
2. `yarn server`
3. [Add sample data](https://github.com/grafana/opensearch-datasource/blob/main/CONTRIBUTING.md#add-sample-data)
4. Run the queries below using the default provisioned data source

Query
---
query

Result
---
table

Explore
---
explore

### DIFF
--- a/pkg/opensearch/response_parser.go
+++ b/pkg/opensearch/response_parser.go
@@ -1122,8 +1122,20 @@ func (rp *responseParser) processAggregationDocs(esAgg *simplejson.Json, aggDef 
 						value = castToFloat(bucket.GetPath(metric.ID, statName))
 					}
 
-					fields = addMetricValue(fields, rp.getMetricName(metric.Type), value)
-					break
+					fieldName := fmt.Sprintf("%v %v", rp.getMetricName(metric.Type), rp.getMetricName(statName))
+					fields = addMetricValue(fields, fieldName, value)
+				}
+			case percentilesType:
+				percentiles := bucket.GetPath(metric.ID, "values")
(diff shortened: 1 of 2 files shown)


idastambuk (approved):
👏🏻

PR #718, opened 2025-07-23
PPL query editor: Add a subset of functions and commands for autocompletion

idastambuk:
**What is this feature?**

This adds PPL suggestion support for Opensearch commands:
```
search
where
fields
stats
eventstats
dedup
sort
eval
head
top
rare
parse
```

 1. The first commit is the language support. 
 - The docs to support this are [here](https://github.com/opensearch-project/sql/blob/main/ppl/src/main/antlr/OpenSearchPPLParser.g4)
-  A lot was taken from this [PR](https://github.com/grafana/grafana-security-mirror/pull/271). After reviewing the differences between the docs for that and OpenSearch PPL, I made some small changes (`top` command, `eval` command) and added the `search` command suggestions
- Also added field lookup for index fields. I memoized that function so that the call isn't made every time fields are suggested (This also improves the field dropdown in Lucene editor.) 
A future improvement would be to also fetch index names when `search index=/source=` is used, then fetch fields depending on the selected index
2. The second commit is the tests

**Testing:** 
There are some sample queries here: [PPLSQL Sample Queries.docx](https://github.com/user-attachments/files/17801174/PPLSQL.Sample.Queries.docx) from the cloudwatch feature, excluding search command, of which there are examples [here](https://docs.opensearch.org/docs/latest/search-plugins/sql/ppl/index/)

-->

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -275,7 +275,8 @@ export const convertOrderByToMetricId = (orderBy: string): string | undefined =>
 // memoizes results of async calls based on the arguments passed to the function
 export function memoizeAsync<Args extends unknown[], Result>(
   fn: (...args: Args) => Promise<Result>,
-  getKey: (...args: Args) => string = (...args) => JSON.stringify(args)
+  getKey: (...args: Args) => string = (...args) => JSON.stringify(args),
+  maxSize = 10
 ): (...args: Args) => Promise<Result> {
   const cache = new Map<string, Result>();
 
@@ -289,6 +290,15 @@ export function memoizeAsync<Args extends unknown[], Result>(
 
     const promise = fn(...args).then((result) => {
       cache.set(key, result);
+
+      // Limit cache size
+      if (cache.size > maxSize) {
+        const firstKey = cache.keys().next().value; // delete first item in the cache
+        if (firstKey) {
+          cache.delete(firstKey);
+        }
+      }
+
       return result;
     });
 
--- a/src/utils.test.ts
+++ b/src/utils.test.ts
@@ -145,4 +145,28 @@ describe('memoizeAsync', () => {
       expect(spy).toHaveBeenCalledTimes(2);
     });
   });
+  it('should delete first item in the cache when the cache size exceeds the limit', async () => {
+    const spy = jest.fn(async (x: number) => {
+      return x * 2;
+    });
+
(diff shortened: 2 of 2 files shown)


src/utils.ts
```diff
@@ -271,3 +271,28 @@ export const convertOrderByToMetricId = (orderBy: string): string | undefined =>
...
+  getKey: (...args: Args) => string = (...args) => JSON.stringify(args)
+): (...args: Args) => Promise<Result> {
+  const cache = new Map<string, Promise<Result>>();
+
+  return (...args: Args): Promise<Result> => {
+    const key = getKey(...args);
+
+    if (cache.has(key)) {
+      return cache.get(key)!;
+    }
```
kevinwcyu, line 286: to get rid of the non-null assertion

```suggestion
    const cachedValue = cache.get(key);
    if (cachedValue) {
      return cachedValue;
    }
```

src/utils.ts
```diff
@@ -271,3 +271,28 @@ export const convertOrderByToMetricId = (orderBy: string): string | undefined =>
...
+
+    const resultPromise = fn(...args);
+    cache.set(key, resultPromise);
```
kevinwcyu, line 289: caching the promise may leak memory. how about we cache the returned value instead of the promise?

idastambuk: Good call, changed so it caches the result

src/opensearchDatasource.ts
```diff
@@ -704,9 +707,8 @@ export class OpenSearchDatasource
   isMetadataField(fieldName: string) {
     return META_FIELDS.includes(fieldName);
   }
-
   // TODO: instead of being a string, this could be a custom type representing all the available types
-  async getFields(type?: string, range?: TimeRange): Promise<MetricFindValue[]> {
+  getFields = memoizeAsync(async (type?: string, range?: TimeRange): Promise<MetricFindValue[]> => {
```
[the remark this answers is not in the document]
idastambuk: Yes theoretically it might return outdated results when constructing dynamic patterns. 
I don't think we should worry too much about this particular call being cached though, since indices are more likely to be stored on a daily pattern than minute or seconds. Also, fields are not normally added/removed from indices often. Caching with `defaultTimeRange` would be difficult since it's a relative range `now-6h` so I'm not sure it's worth figuring out how to cache this, unless you feel strongly about finding a workaround!

PR #772, opened 2025-09-02
PPL Language Support part 2

idastambuk:
This adds language (syntax highlighting and suggestions) support for the remainder of PPL commands: 
joinCommand

- renameCommand
- statsCommand
- grokCommand
- patternsCommand
- lookupCommand
- kmeansCommand
- adCommand
- mlCommand
- fillnullCommand
- trendlineCommand
- appendcolCommand
- expandCommand
- flattenCommand
- reverseCommand

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "@types/prismjs": "1.26.5",
     "@types/react": "^18.3.12",
     "@types/react-test-renderer": "19.1.0",
-    "@types/semver": "^7.7.0",
+    "@types/semver": "^7.7.1",
     "@typescript-eslint/eslint-plugin": "^8.38.0",
     "@typescript-eslint/parser": "^8.40.0",
     "copy-webpack-plugin": "^13.0.1",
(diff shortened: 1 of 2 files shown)


src/__mocks__/ppl-test-data/singleLineQueries.ts
```diff
@@ -479,3 +479,372 @@ export const queryWithFunctionCalls = {
     ],
   ] as monacoTypes.Token[][],
 };
+
+export const joinQuery = {
+  query:
+    'left outer join left = table1 right = table2 left_hint.id = leftIdent on isnull(leftIdent) = false right = table2',
```
kevinwcyu, line 485: is `left` supposed to be suggested? when i type `left` i don't get that as a suggestion, it also isn't followed by `outer` as a suggestion.

kevinwcyu (approved):
looks good, just wondering about the join query suggestions in the comment below.

src/__mocks__/ppl-test-data/singleLineQueries.ts
idastambuk: I left that out of the suggestions cause I thought it could be confusing bundling it with the commands. But it makes sense to suggest outer and join after left/right, will add that

PR #1095, opened 2026-05-27
security hardening

kevinwcyu:
## Changes

- Ran @grafana/create-plugin update to get the .config/ directory files up-to-date

- Hardened Yarn supply-chain settings:
  - Upgraded pinned Yarn from `4.13.0` to `4.15.0`.
  - Migrated `yarn.lock` metadata to Yarn 4.15 lockfile version `10`.
  - Enabled immutable installs, hardened mode, strict checksum behavior, exact semver defaults, npm registry pinning, and HTTP source blocking.
  - Kept lifecycle scripts disabled and the 3-day npm minimum age gate with `@grafana/*` preapproval.

- Reduced direct npm dependency surface:
  - Removed unused/duplicate direct dependencies from `package.json`.
  - Converted ranged direct dependencies to exact versions.
  - Marked the package as private to prevent accidental npm publication.
  - Pinned the plugin signing command instead of using `@latest`.

- Hardened GitHub Actions:
  - Replaced mutable checkout/install of `grafana/grafana-github-actions` with a pinned `grafana/grafana-github-actions/commands` action.
  - Kept upstream’s `create-github-app-token` token flow.
  - Added `--mode=skip-build` to the local Yarn install step.
  - Pinned the duplicate `github/combine-prs` workflow reference.
  - Tightened Zizmor config by removing the broad `github/*` unpinned-action exception.
  - Updated `grafana/plugin-ci-workflows` CI/CD reusable workflows from `v7.3.1` to `v8.0.1`.

- Updated Go security posture:
  - Updated Go from `1.25.8` to `1.26.3`.
  - Updated `golang.org/x/net` to `v0.55.0` and related `golang.org/x/*` modules.
  - Removed the stale `golang.org/x/sys` replace directive so the newer resolved version is used.

- Added TruffleHog false-positive handling:
  - Added `.trufflehogignore` for generated Yarn release bundles matching `.yarn/releases/yarn-x.x.x.cjs`.
  - This addresses the unverified GitLab-token false positive reported in the generated Yarn bundle while keeping the ignore scoped to Yarn release artifacts.

- Removed obsolete repository config:
  - Removed duplicate/unused GitHub config files and workflows superseded by current workflow locations.

## Validation

- `yarn install --immutable --mode=skip-build`
- `yarn test:ci`
- `yarn typecheck`
- `go test ./...`
- `govulncheck ./...` reports 0 reachable vulnerabilities
- npm/Yarn supply-chain audit passes all checks


.github/workflows/issue_commands.yml
```diff
@@ -11,22 +11,13 @@ jobs:
...
-      - name: Install Actions
-        run: npm install --production --prefix ./actions
       - name: Get GitHub App token
         id: get-github-app-token
         uses: grafana/shared-workflows/actions/create-github-app-token@259ba21cb3ff07724f331e26d926d655d24b317b # create-github-app-token/v0.2.3
         with:
           github_app: aws-ds-token-creator
       - name: Run Commands
-        uses: ./actions/commands
+        uses: grafana/grafana-github-actions/commands@3c62d35c5a66e730186a338e45aeb8fa784e2d56
```
aangelisc, line 20: Does this work?

aangelisc (approved):
Just one non-blocking q, otherwise LGTM

.github/workflows/issue_commands.yml
kevinwcyu: i believe it should work based on this https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-using-a-public-action-in-a-subdirectory

PR #1136, opened 2026-07-20
Fix Logs Volume supplementary queries for PPL log queries

haarisshh10:
**What this PR does / why we need it**

Previously, Logs Volume supplementary queries were only generated for queries where metrics[0]. type was logs.

PPL log queries use the format: "logs" with metrics[0]. Type is set to count, causing them to be treated as unsuitable. As a result, the supplementary Logs Volume query was not generated, and the Logs Volume histogram was not displayed in Explore.

This change updates the suitability check also to recognize PPL log queries (queryType === "PPL" and format === "logs"), allowing the supplementary Logs Volume query to be generated while preserving the existing behavior for other query types.

Which issue(s) does this PR fix
**Fixes #1057**

**Manually verified the following scenarios:**

- PPL + Logs: The Logs Volume histogram is displayed.
- Lucene + Logs: Existing behavior remains unchanged.
- PPL + Table: No supplementary Logs Volume query is generated.
- Lucene + Metric: Existing behavior remains unchanged.


haarisshh10:
image

src/opensearchDatasource.ts
```diff
@@ -195,7 +195,12 @@ export class OpenSearchDatasource
     switch (options.type) {
       case SupplementaryQueryType.LogsVolume:
         // it has to be a logs-producing range-query
-        isQuerySuitable = !!(query.metrics?.length === 1 && query.metrics[0].type === 'logs');
+        isQuerySuitable =
```
iwysiu: This correctly recognizes PPL logs queries. The problem is what we return below: we still build a Lucene count + `date_histogram` query while copying `query.query` verbatim and omitting `queryType` (backend defaults to Lucene). That turns a PPL string like `source = idx | where …` into a Lucene `query_string` filter, so volume can render but won’t match the log query. Can we return a PPL-shaped volume query (or an explicit Lucene query with a proper filter/index) instead of reusing the PPL text as Lucene?

iwysiu (commented):
Hi @haarisshh10 , thanks for the pr!

This doesn't fully handle the issue, now the query passes through the isQuerySuitable check but we're still building a lucene query either way, while we want to build a PPL query if that's the type.

Also, can you add tests for this?

haarisshh10:
Hi @iwysiu 

Thanks for the review! I understand the issue now. While investigating, I found that getSupplementaryQuery() always constructs a Lucene-style aggregation (count + date_histogram), and I couldn't find an existing implementation in the datasource for generating a PPL logs volume query.

Would you prefer that the supplementary query preserve the original PPL query and metadata, or should it instead construct a new PPL query (for example, using stats/time bucketing) specifically for the logs volume request?

iwysiu:
Hi @haarisshh10 — thanks for digging into this!

We should **construct a new PPL logs-volume query**, not reuse the original PPL string on the Lucene count + `date_histogram` path. Keep the user’s `source` / filters, then aggregate counts into time buckets (same idea as the Lucene supplementary query).

Please use **`stats` + `span()`**, e.g. `| stats count() by span(@timestamp, 5m)` — see https://docs.opensearch.org/latest/sql-and-ppl/ppl/commands/stats/

That pattern has been available since early PPL; we shouldn’t rely on newer commands that aren’t present on all OpenSearch versions we support.

If a log level field is configured, also group by that field so the histogram can still split by level, similar to the Lucene path.

Please add tests for PPL logs (volume query generated + shape), and negative cases (PPL table / non-logs still skip supplementary volume).

Happy to iterate on the exact query shape once you have a first pass!

haarisshh10:
Hi @iwysiu 

Thanks for the review! I've updated the implementation based on the feedback:

Added a dedicated PPL logs volume builder instead of using the Lucene aggregation path.

The supplementary query now preserves the original PPL query and appends a stats count() by span(...) aggregation.

Added regression tests for PPL logs and PPL table queries.

Manually verified that PPL logs generate the supplementary log-volume-* query and that PPL table queries do not.

src/opensearchDatasource.ts
```diff
@@ -190,56 +190,80 @@ export class OpenSearchDatasource
       return undefined;
     }
 
-    let isQuerySuitable = false;
+    const isLuceneLogQuery = query.metrics?.length === 1 && query.metrics[0].type === 'logs';
+    const isPPLLogQuery = query.queryType === QueryType.PPL && query.format === 'logs';
```
iwysiu: `isLuceneLogQuery` doesn't exclude PPL. After a Lucene→PPL switch the editor often leaves `metrics: [{ type: 'logs' }]`, so PPL+table still passes the gate and hits `buildPPLLogsVolumeQuery`. Suggest branching on query type first:

```suggestion
    const isPPL = query.queryType === QueryType.PPL;
    const isLuceneLogQuery = !isPPL && query.metrics?.length === 1 && query.metrics[0].type === 'logs';
    const isPPLLogQuery = isPPL && query.format === 'logs';
```

src/opensearchDatasource.ts
```diff
@@ -190,56 +190,80 @@ export class OpenSearchDatasource
...
+    const logLevelField = this.logLevelField ? `, \`${this.logLevelField}\`` : '';
+    const statsClause = `| stats count() by span(\`${timeField}\`, \$__interval)${logLevelField}`;
+
+    return {
+      ...query,
+      refId: `${REF_ID_STARTER_LOG_VOLUME}${query.refId}`,
+      query: [queryText, statsClause].filter(Boolean).join('\n'),
+      queryType: QueryType.PPL,
+    };
+  }
```
iwysiu, line 273: Spreading `...query` keeps `format: 'logs'`. Backend `parseResponse` then uses `parseLogs`, so the volume frame gets `PreferredVisualization: logs` instead of a time series. Please set `format: 'time_series'` explicitly (and avoid spreading fields the volume path doesn't need).

src/opensearchDatasource.ts
```diff
@@ -190,56 +190,80 @@ export class OpenSearchDatasource
...
+      timeField,
+      bucketAggs,
+    };
+  }
+
+  private buildPPLLogsVolumeQuery(query: OpenSearchQuery): OpenSearchQuery {
+    const timeField = this.timeField ?? '@timestamp';
+    const queryText = query.query?.trim() || '';
+    const logLevelField = this.logLevelField ? `, \`${this.logLevelField}\`` : '';
+    const statsClause = `| stats count() by span(\`${timeField}\`, \$__interval)${logLevelField}`;
```
iwysiu: When `logLevelField` is set this becomes `| stats count() by span(...), \`level\`` (3 columns). `getTimeSeriesResponseMeta` rejects that with "response should have 2 fields but found 3". For this PR, drop the level split on the PPL path (unsplit histogram is still the fix for #1057), or extend the parser — happy either way.

src/opensearchDatasource.ts
```diff
@@ -190,56 +190,80 @@ export class OpenSearchDatasource
...
+      settings: {
+        interval: 'auto',
+        min_doc_count: '0',
+        trimEdges: '0',
+      },
+      field: timeField,
+    });
+
+    return {
+      ...query,
```
iwysiu: The old path returned a clean object (`refId`, `query`, `metrics`, `timeField`, `bucketAggs`). Spreading `...query` now drags `format` / `luceneQueryType` / etc. into the volume query. Can we keep the previous shape?

src/opensearchDatasource.test.ts
```diff
@@ -1103,6 +1103,40 @@ describe('OpenSearchDatasource', function (this: any) {
...
+      expect(
+        ds.getSupplementaryQuery(
+          { type: SupplementaryQueryType.LogsVolume },
+          {
+            refId: 'A',
+            query: 'source = logs | stats count() by host',
+            queryType: QueryType.PPL,
+            format: 'table',
+          }
+        )
```
iwysiu, line 1155: These won't catch the issues above: no `format` assertion, PPL-table fixture omits leftover `metrics`, and no `logLevelField` case. Worth asserting the full result object (including `format: 'time_series'`) plus those negatives.

iwysiu (changes requested):
Thanks for the update — dedicated PPL builder + `stats`/`span` is the right direction. A few issues still block a usable Logs Volume histogram:

**Must fix:**
- `buildPPLLogsVolumeQuery` spreads the original query and never sets `format: 'time_series'`, so the backend routes the `stats`/`span` response through `parseLogs` instead of `parseTimeSeries`.
- Grouping by `logLevelField` produces 3 columns; the PPL time-series parser only accepts 2 (`response should have 2 fields but found 3`). Prefer dropping the level split for PPL in this PR, or extending the parser.
- Suitability check: `isLuceneLogQuery` doesn't exclude PPL, so a PPL+table query that still carries leftover Lucene `metrics: [{ type: 'logs' }]` incorrectly gets a volume query. Branch on `queryType` first.

**Should fix:**
- Don't `...query` in `buildLuceneLogsVolumeQuery` — keep the previous clean return shape.
- `$__interval` isn't substituted on the PPL backend path the way Lucene batch encoding does; consider deriving the span from the request interval (and skip empty query text).
- Tighten tests: assert full query/`format`, leftover-metrics PPL table negative case, and (if kept) logLevelField behavior.

Verified the format and 3-column parser behaviors against the current PPL response parser.

haarisshh10:
Hello @iwysiu 

Thanks for the detailed review! I've addressed the requested implementation changes around the suitability check, format: "time_series", removing the PPL log-level split, and returning clean supplementary query objects. I also strengthened the regression tests.

Regarding the interval comment, I investigated it before updating further. In my local testing, the generated PPL supplementary query is sent to /api/ds/query with the interval already resolved (for example span(@timestamp, 5s) rather than span(@timestamp, $__interval)), so I wasn't able to reproduce $__interval reaching the backend unchanged. If there's a specific scenario where interpolation doesn't happen, I'd be happy to adjust the implementation to derive the span from the request interval instead.

src/opensearchDatasource.test.ts
```diff
@@ -1103,6 +1103,43 @@ describe('OpenSearchDatasource', function (this: any) {
...
+          format: 'logs',
+        }
+      );
+
+      expect(result).toMatchObject({
+        refId: 'log-volume-A',
+        queryType: QueryType.PPL,
+        format: 'time_series',
+      });
+      expect(result?.query).toContain('stats count()');
```
iwysiu: Nit: You know what the expected query is so you could just test this as an equals

iwysiu (commented):
Hi @haarisshh10 ! Thanks for the edits. I looked into the `$__interval`  logic and I saw the same thing.

Almost approved! I have one optional testing note, but before I can merge this I need you to sign both of these commits, its [required for the repo](https://github.com/grafana/opensearch-datasource/blob/main/CONTRIBUTING.md#signed-commits-are-required)

iwysiu (commented):
@haarisshh10 Thanks for work!
One last thing: I ran the tests and `returns logs volume query for PPL logs` expects pipes rewritten onto separate lines, but the builder only appends `\n| stats…`, can you update the assertion to match?

iwysiu (dismissed):

iwysiu:
@haarisshh10 Github is still saying that not all the commits are verified, can you double check that?

haarisshh10:
Hi @iwysiu 
Thanks! I've updated the branch, and all commits are now verified. Please let me know if there's anything else I should address.

iwysiu (approved):

PR #1137, opened 2026-07-20
Fix PPL autocomplete for source, indices, pipe, and field values

### LINKED ISSUE #872: PPL Auto-complete issues
**What happened**:
PPL auto-complete doesn't work as expected. When typing so as the first character, source is not listed. When source is typed, indices do not show up in the list, instead fields are shown. When a field name is selected in. a where clause, values don't show up. Recommendations for PPL commands are not placement appropriate.

Auto-complete experience as present in OpenSearch Discover

1. Go to Explore, select an OpenSearch datasource.
2. Select PPL as query type
3. Type s in the query editor, source doesn't show up. Type source=sample_app_logs, pipe symbol won't be prompted. Type in a where clause with a field name, field values aren't prompted.

**Screenshots**

Image

Image

Image

 Expected:

Image

Image

Image

Image

**Environment**:

- Grafana version: 10.4.1
- Plugin version: 2.31.1
- OS Grafana is installed on: Amazon Managed Grafana
- User OS & Browser: MacOS, Chrome
- Others:

iwysiu:
## Summary
- Fixes [#872](https://github.com/grafana/opensearch-datasource/issues/872) by aligning PPL autocomplete with Discover-style editing: suggest `source`/`index` at query start, indices after `source =`/`index =`, `|` after a completed source clause, and field values after comparisons in `where`/`search`.
- Wires existing `getIndices()` / `getTerms()` into `PPLCompletionItemProvider` and adds statement positions (`StartOfQuery`, `AfterFromClause`, `AfterFromClauseComplete`, `AfterComparisonOperator`).
- Lefthook: drop eslint `--ignore-path` (unsupported with flat config). No hardcoded `node_modules/.bin` paths — after `yarn install`, `yarn prettier` / `yarn cspell` resolve correctly from the lockfile.

## Test plan
- [ ] `yarn install` (if local node_modules is stale after the Yarn 4 upgrade)
- [ ] Unit: `./node_modules/.bin/jest --watchAll=false src/language/ppl/completion/statementPosition.test.ts src/language/ppl/completion/PPLCompletionItemProvider.test.ts`
- [ ] Explore → OpenSearch → PPL: empty editor suggests `source` when typing `s`
- [ ] After `source = `, suggestions are index names (not fields)
- [ ] After `source = <index> `, `|` is suggested
- [ ] After `where <field> = `, field values appear
- [ ] After `|`, commands still suggest (without re-prompting `source =` as the primary start flow)
- [ ] Pre-commit hook: stage a `.ts` file and confirm `yarn prettier` / `yarn cspell` / eslint succeed


iwysiu:
### Follow-up (out of scope for this PR)

When testing `source = opensearch_dashboards_sample_data_logs | fields `, field suggestions can include names that are not valid for that source index.

**Why:** PPL completion still calls `getFields()` → `_field_caps` against the **datasource-configured** index pattern (`jsonData.database`), not the index from `source=` / `index=` in the query text. This is pre-existing; this PR only improves source/index/pipe/value suggestions.

**Follow-up work:** Parse `source=`/`index=` from the current PPL query and scope field (and likely term) suggestions to that index. Optionally also filter multi-field noise (e.g. `.keyword`) / nested object paths to better match Discover.

itsmylife (changes requested):
I did some manual testing. (I let agent do the test first and then I did manual testing once more)
Here are my findings. 

1. Pipe is not suggested 
When you type `source = <something with - in it>` e.g. `source = logs-2026` and then put a space nothing happens. If I do it something which has no dash then it suggests. I'm not sure if I'm suppose to put that value in quotes or backticks.

2. field with dash breaks suggestion
this works 
```
source = opensearch_dashboards_sample_data_logs | where user-id = 'configured-user'
```

but if I delete `'configured-user'` and let autocomplete suggests me I got nothing.

3. wrong suggestion (this is coming from agent directly)

When you type `source = logs-2024 | where status = `
it suggests `'configured-value'` but `logs-2024` document contains: `source-value` so it should suggest that. 

4. When you leave the codeeditor in autocomplete mode (like clicking outside etc) and return back to codeeditor, the autocomplete list very briefly showed up and then returns to `Loading ...` state and stays that way. I had to press esc and then trigger auto completion.

iwysiu:
Thanks for the thorough manual testing — addressed in f8388d1:

1. **Pipe after hyphenated index** (`source = logs-2026 `) — Monarch splits on `-`, so we now join those segments when detecting a completed from-clause and suggest `|`.
2. **Values for hyphenated fields** (`where user-id = `) — same join logic when resolving the field name for `getTerms` (was previously looking up only `id`).
3. **Wrong term values after `source = logs-2024`** — still intentionally out of scope here; terms/fields are scoped to the query `source=`/`index=` in the stacked follow-up #1156.
4. **Stuck on `Loading...` after blur/refocus** — suggest was firing while the completion provider was being disposed/re-registered. We now trigger suggest only after re-registration completes.

Would appreciate another pass when you have a chance.

itsmylife:
I did an agentic code review followed by manual browser testing against `f8388d1` in the local devenv.

The follow-up fixes are working:
- `source = logs-2024 ` now suggests `|`.
- `where user-id = ` now produces term suggestions.
- Blur/refocus no longer leaves autocomplete stuck on `Loading...`.

Two correctness issues remain:

1. **Term suggestions still use the datasource-configured index instead of the query source.**

   With `logs-2024.status = source-value` and the configured index containing `status = configured-value`, this query:

   ```ppl
   source = logs-2024 | where status = 
   ```

   suggests `'configured-value'`. The autocomplete `_msearch` request confirms that its header uses:

   ```json
   { "index": "opensearch_dashboards_sample_data_logs" }
   ```

   Since this PR introduces field-value autocomplete and #872 explicitly asks for it, I don't consider correctness of those values out of scope. Please fold the source-scoping fix from #1156 into this PR, or remove field-value completion from this PR's scope.

2. **Fields named `index` or `source` are misclassified as from-clauses.**

   Both of these queries suggest index names rather than values for the field:

   ```ppl
   source = logs-2024 | where index = 
   source = logs-2024 | where source = 
   ```

   Only query-start `source =` / `index =` (or those immediately after `search`) should enter index-name completion.

I verified both failures manually in Explore, in addition to tracing the relevant code paths. Keeping my changes-requested verdict until these are addressed.

Disclaimer: This text was generated via an agent.

iwysiu:
Thanks for re-testing. Addressed on the branch:

1. Folded source-scoped field/term autocomplete from #1156 into this PR so values use the query `source=` / `index=`.
2. Fixed `where index =` / `where source =` misclassified as from-clauses — only true from-clauses (`source=` / `index=` at query start or after `search`) enter index-name completion. (`index=` is a valid synonym for `source=` in the OpenSearch PPL parser.)

Would appreciate another pass when you have a chance.

itsmylife (approved):
LGTM 🚀 
This is still happening "When you leave the codeeditor in autocomplete mode (like clicking outside etc) and return back to codeeditor, the autocomplete list very briefly showed up and then returns to Loading ... state and stays that way. I had to press esc and then trigger auto completion." But I won't block PR. it can be addressed in another PR.

PR #1158, opened 2026-07-30
Fixed Auto interval bug

### LINKED ISSUE #1123: 'Auto' interval breaking with nested aggregations
**What's the problem?**
When a panel uses Auto interval on a date histogram with a nested Terms Group By, the OpenSearch plugin computes the interval purely from time range ÷ (panel width / maxDataPoints). It has no awareness that total buckets = time buckets × terms cardinality. The result is that dashboards which render fine for one time range silently fail for another with the OpenSearch error "Trying to create too many buckets. Must be less than or equal to: [65535]". User has confirmed there is no static timeInterval floor that is safe across all combinations: because both time range and cardinality vary at query time, any fixed floor will still be exceeded by some range/cardinality pairing. OpenSearch Dashboards (Kibana) handles the same query without error by back-calculating a safe interval.

**Proposed solution:**
When Auto interval is selected, the plugin should compute a safe interval floor at query time by accounting for the bucket multiplication of nested Terms aggregations — i.e. back-calculate from search.max_buckets ÷ estimated terms cardinality before sending the query. OpenSearch's native `auto_date_histogram` aggregation already implements this behavior (it caps returned buckets at or below a target count and automatically widens the interval when data would exceed it). Evaluating whether the plugin can adopt `auto_date_histogram`, or replicate its back-calculation, is the likely implementation path. This would fix all existing and future dashboards simultaneously with zero customer action.

Extension of ticket https://github.com/grafana/opensearch-datasource/issues/426

mattvella07:

When a date_histogram used the "Auto" interval alongside nested Terms
group-by aggregations, the interval was computed from the time range
alone. Total buckets are time buckets x terms sizes, so queries could
exceed OpenSearch's search.max_buckets limit and fail with
"Trying to create too many buckets".

- Make the auto-interval calculation account for the terms multiplier
  by budgeting total buckets and raising the interval when needed.
- Round the required interval up to a clean, representable value so it
  is never truncated below the budget.
- Only affects date histograms using "Auto"; queries without terms
  aggregations are unchanged.
- Add unit tests for the interval math and a snapshot regression test.

Fixes #1123

Before 

before

After 

after


pkg/opensearch/lucene_handler.go
```diff
@@ -110,6 +117,42 @@ func (h *luceneHandler) processQuery(q *Query) error {
...
+		if bucketAgg.Type != termsType {
+			continue
+		}
+		size := 500
+		if s, err := bucketAgg.Settings.Get("size").Int(); err == nil {
+			size = s
+		} else if s, err := bucketAgg.Settings.Get("size").String(); err == nil {
+			size, err = strconv.Atoi(s)
+			if err != nil {
+				size = 500
```
codyconfer: I think these are values we are supposed to react to, not assert.

```go
func termsBucketProduct(bucketAggs []*BucketAgg, ceiling int64) int64 {
      var product int64 = 1
      for _, bucketAgg := range bucketAggs {
              if bucketAgg.Type != termsType {
                      continue
              }
              size, ok := configuredTermsSize(bucketAgg)
              if !ok {
                      continue
              }
              product *= int64(size)
              if product >= ceiling {
                      return ceiling
              }
      }
      return product

func configuredTermsSize(bucketAgg *BucketAgg) (int, bool) {
      if s, err := bucketAgg.Settings.Get("size").Int(); err == nil {
              return s, s > 0
      }
      if s, err := bucketAgg.Settings.Get("size").String(); err == nil {
              if n, err := strconv.Atoi(s); err == nil {
                      return n, n > 0
              }
      }
      return 0, false
}
```

pkg/opensearch/lucene_handler.go
```diff
@@ -55,7 +55,14 @@ func (h *luceneHandler) processQuery(q *Query) error {
 	if err != nil {
 		return err
 	}
-	interval := tsdb.CalculateInterval(&h.reqQueries[0].TimeRange, minInterval)
+	// Only date histograms with an "auto" interval are affected by the bucket
+	// budget; when combined with terms aggregations the total bucket count can
+	// exceed OpenSearch's search.max_buckets, so pass the terms product along.
```
codyconfer, line 64:
```go
maxBuckets := tsdb.MaxBucketsFrom(h.dsSettings)
termsProduct := int64(1)
if hasAutoDateHistogram(q.BucketAggs) {
      termsProduct = termsBucketProduct(q.BucketAggs, maxBuckets)
}
interval, err := tsdb.CalculateInterval(&h.reqQueries[0].TimeRange, minInterval, termsProduct, maxBuckets)
if err != nil {
       return backend.DownstreamError(err)
}
if interval.RaisedForBuckets {
      q.intervalRaisedTo = interval.Text
}
```

pkg/tsdb/interval.go
```diff
@@ -24,19 +30,98 @@ func (i *Interval) Milliseconds() int64 {
...
+	if rounded < effectiveMin {
+		rounded = effectiveMin
+	}
 	return Interval{Text: FormatDuration(rounded), Value: rounded}
 }
 
+// bucketFloorInterval computes the smallest date histogram interval that keeps the
+// total bucket count (time buckets multiplied by termsProduct) under maxBucketsTarget.
+// It returns 0 when termsProduct <= 1 so behavior is unchanged for queries without
+// terms aggregations.
```
codyconfer:
```go
func bucketFloorInterval(span, termsProduct, maxBuckets int64) (time.Duration, error) {
      if termsProduct <= 1 {
              return 0, nil
      }

      budget := maxBuckets * bucketHeadroomPercent / 100
      timeTarget := budget / termsProduct
      if timeTarget < minTimeBuckets {
              return 0, fmt.Errorf(
                      "%w: terms aggregations would produce up to %d buckets per time bucket, "+
                              "leaving fewer than %d time buckets within the %d bucket limit; "+
                              "reduce the terms size or narrow the time range",
                      ErrBucketBudgetOutOfBounds, termsProduct, minTimeBuckets, maxBuckets)
      }

      required := (span + timeTarget - 1) / timeTarget
      return roundIntervalUp(time.Duration(required)), nil
}
```

pkg/tsdb/interval.go
```diff
@@ -24,19 +30,98 @@ func (i *Interval) Milliseconds() int64 {
 	return i.Value.Nanoseconds() / int64(time.Millisecond)
 }
 
-func CalculateInterval(timeRange *backend.TimeRange, minInterval time.Duration) Interval {
+func CalculateInterval(timeRange *backend.TimeRange, minInterval time.Duration, termsProduct int64) Interval {
```
codyconfer:
```go
func CalculateInterval(timeRange *backend.TimeRange, minInterval time.Duration, termsProduct, maxBuckets int64) (Interval, error) {
      span := timeRange.To.UnixNano() - timeRange.From.UnixNano()

      floor, err := bucketFloorInterval(span, termsProduct, maxBuckets)
      if err != nil {
              return Interval{}, err
      }

      effectiveMin := minInterval
      if floor > effectiveMin {
              effectiveMin = floor
      }

      natural := resolveInterval(span, minInterval)
      budgeted := resolveInterval(span, effectiveMin)

      return Interval{
              Text:             FormatDuration(budgeted),
              Value:            budgeted,
              RaisedForBuckets: budgeted > natural,
      }, nil
}

func resolveInterval(span int64, min time.Duration) time.Duration {
      interval := time.Duration(span / defaultRes)
      if interval < min {
              return min
      }
      if rounded := roundInterval(interval); rounded >= min {
              return rounded
      }
      return min
}
```

pkg/tsdb/interval.go
```diff
@@ -15,6 +15,12 @@ var (
 	day              = time.Hour * 24
 )
 
+// maxBucketsTarget is the number of date histogram buckets we aim to stay under.
```
codyconfer: this is marked as a dynamic field in open search, so a constant might not be the right call. 
this would grant headroom to a cluster with default configurations, but if a cluster is configured to search.max_buckets=10000, assuming 60000 will always blow the budget. We should probably have his configurable on release.

pkg/tsdb/interval.go
```diff
@@ -15,6 +15,12 @@ var (
...
+// OpenSearch's default search.max_buckets is 65535; we target below it to leave
+// headroom for parent-level and extended_bounds buckets. This could later be made
+// configurable via datasource settings.
+const maxBucketsTarget int64 = 60000
+
 type Interval struct {
 	Text  string
 	Value time.Duration
```
codyconfer, line 39:
```go
RaisedForBuckets bool
```

pkg/tsdb/interval.go
```diff
@@ -15,6 +15,12 @@ var (
 	day              = time.Hour * 24
 )
 
+// maxBucketsTarget is the number of date histogram buckets we aim to stay under.
+// OpenSearch's default search.max_buckets is 65535; we target below it to leave
+// headroom for parent-level and extended_bounds buckets. This could later be made
+// configurable via datasource settings.
+const maxBucketsTarget int64 = 60000
```
codyconfer:
```go
const (
    // Matches open search default
    defaultMaxBuckets int64 = 65535
    // Our calculated budget
    bucketHeadroomPercent int64 = 90
    // Don't collapse small time series
    minTimeBuckets int64 = 20
)
```

pkg/tsdb/interval.go
```diff
@@ -15,6 +15,12 @@ var (
...
+
```
codyconfer, line 36:
```go
// ErrBucketBudgetOutOfBounds means no interval can keep the query under the
// bucket limit while still returning a usable series.
var ErrBucketBudgetOutOfBounds = errors.New("bucket budget out of bounds")
```

pkg/tsdb/interval.go
```diff
@@ -24,19 +30,98 @@ func (i *Interval) Milliseconds() int64 {
 	return i.Value.Nanoseconds() / int64(time.Millisecond)
 }
```
codyconfer, line 48:
```go
func MaxBucketsFrom(dsInfo *backend.DataSourceInstanceSettings) int64 {
      if dsInfo == nil {
              return defaultMaxBuckets
      }
      jsonData, err := simplejson.NewJson([]byte(dsInfo.JSONData))
      if err != nil {
              return defaultMaxBuckets
      }
      if v, err := jsonData.Get("maxBuckets").Int64(); err == nil && v > 0 {
              return v
      }
      return defaultMaxBuckets
}
```

codyconfer (changes requested):
Two items here, and a bunch of code you can ask claude to implement...
1. I think size is a setting that we need to be more reactive in handling rather than asserting our own defaults. So the important changes are in this [comment](https://github.com/grafana/opensearch-datasource/pull/1158/changes#r3694231865).
2. max_buckets is marked as a dynamic field, so we probably shouldn't ship with an overriding constant. it creates headroom at default settings, but will calculate over budget if a cluster is configured under our value. The rest of the source comments are for a calculated budget and surfacing errors to users. 
 
Feel free to ignore point 2 (hopefully the comments are enough for claude to one shot it), but I think the reactive size changes are more necessary.

_in response_parser to surface errors:_
```go
func (rp *responseParser) noteRaisedInterval(frames *data.Frames, target *Query) {
      if target.intervalRaisedTo == "" {
              return
      }
      notice := data.Notice{
              Severity: data.NoticeSeverityInfo,
              Text: fmt.Sprintf(
                      "Interval raised to %s: grouping by terms would exceed the bucket limit at full resolution.",
                      target.intervalRaisedTo),
      }
      for _, frame := range *frames {
              if frame.Meta == nil {
                      frame.Meta = &data.FrameMeta{}
              }
              frame.Meta.Notices = append(frame.Meta.Notices, notice)
      }
}
```

codyconfer (approved):

PR #1168, opened 2026-08-11
test(e2e): add baseline config, query, annotation and variable suites

### LINKED ISSUE #239: Add e2e testing for openSearch
- Use the [instructions](https://github.com/grafana/data-sources/blob/main/docs/testing/e2e-testing.md) to add config and query e2e tests. 
- Follow this [document](https://github.com/grafana/data-sources/blob/main/docs/testing/cloud-e2e-testing.md) to ensure that the e2e tests run nightly against a live Grafana instance
- once the above are completed, ensure the "Data Source Testing" tab is updated in the [spreadsheet](https://docs.google.com/spreadsheets/d/1cwUeUhoKofc2GHGPZlZfaMRoOdkjUGxlQYfek1kqrdg/edit?gid=1011020740#gid=1011020740).

fabrizio-grafana:
Closes #239

Adds a baseline end-to-end suite and wires up the two Cloud test lanes, as per internal test practices.

Tests move from `tests/` to `tests/e2e/`, as the guide prescribes.

### Fixture data

A seeded generator (`seed=42`) produces two indices over a pinned four-hour window, loaded by a one-shot `opensearch-loader` container that Grafana now waits on via `service_completed_successfully` — so tests can never race an empty cluster.

Mappings are explicit rather than dynamic on purpose: `traceId`, `traceGroup` and `serviceName` must be `keyword` for the terms aggregations behind the Traces query type. Under dynamic mapping they become `text` and every trace query silently returns nothing. See `tests/e2e/fixtures/README.md`.

Two provisioned datasources (`OpenSearch E2E Logs`, `OpenSearch E2E Traces`) point at them. The three pre-existing datasources are unchanged apart from their URL, which moves from `host.docker.internal` to the `opensearch-node1` service name so it resolves regardless of Docker flavour.


mattvella07 (approved):
LGTM

PR #1171, opened 2026-08-21
Chore: Use npm as package manager

itsmylife:
## Summary
- Replace Yarn with npm and regenerate the dependency lockfile
- Update workflows, documentation, and development commands to use npm
- Preserve the existing Redux Toolkit 1.9 / Redux 4 dependency stack for Grafana 10.4 compatibility

## Dependency resolution reasoning
Yarn's peer virtualization allowed Redux Toolkit 1.9 to coexist with packages that use newer Redux versions. npm uses a hoisted dependency tree and initially selected `react-redux@9` and Redux 5 at the root, which led to upgrading Redux Toolkit from 1.9.7 to 2.0.0 to satisfy peer dependencies.

The Grafana 10.4 E2E job then failed deterministically after query-editor interactions: selecting **Raw Data** did not update the checked radio, and switching to PPL never mounted the PPL editor. The same tests passed against newer Grafana versions, and the baseline Yarn build passed on Grafana 10.4, isolating the regression to the changed Redux dependency stack rather than Playwright timing.

This PR now keeps Redux Toolkit 1.9.7 and explicitly pins its compatible `react-redux@8.1.3` and `redux@4.2.1` peers. npm nests Redux 5 / React Redux 9 under the Grafana UI dependencies that require them, avoiding the incompatible root-level upgrade.

## Test plan
- [x] `npm ci --ignore-scripts`
- [x] `npm run typecheck`
- [x] `npm run test:ci -- src/components/QueryEditor src/hooks/useStatelessReducer.test.ts` (107 tests)
- [x] `npm run lint -- --no-cache`
- [x] `npm run build`
- [x] Re-run the three failed query-editor scenarios against Grafana Enterprise 10.4.19


src/configuration/DataLink.tsx
akkikumar72, line 5: `LegacyForms` is deprecated.  mentioned by GH workflow. Does it need to be fixed here?

package.json
```diff
@@ -104,7 +105,8 @@
     "webpack-virtual-modules": "0.6.2"
   },
   "engines": {
-    "node": ">=20"
+    "node": ">=24",
+    "npm": ">=11.16.0"
   },
-  "packageManager": "yarn@4.17.1"
+  "packageManager": "npm@11.18.0"
```
akkikumar72: Node 24.19.0 is successfully exercised by the current CI run, so the Node 24 change looks valid. 
However, the setup logs [show npm 11.17.0 ](https://github.com/grafana/opensearch-datasource/actions/runs/32727635867/job/97432470953) while packageManager pins npm 11.18.0. 
Should CI explicitly install 11.18.0, or should the packageManager value match the npm version bundled with Node 24.19.0?
Screenshot 2026-08-25 at 14 20 14

itsmylife: It should match the version. I'll update it.

src/configuration/DataLink.tsx
itsmylife: Outside the scope of this PR. I won't fix it in here.

akkikumar72 (approved):

